Issue #223, opened 2025-07-28
file_converter.py cannot extract the last SCF energy from .log files that contain link jobs


ZihanL1:
[intervening_nonoverlapping_A'_A.log](https://github.com/user-attachments/files/21463258/intervening_nonoverlapping_A._A.log)  When I use "file_converter.py -d . -t log" command to convert log file to xyz file, the energy written in the output .xyz file is not the final SCF energy, but the first one. It seems that file_converter.py may not correctly recognize and extract the final SCF energy when processing log files containing Link1 jobs.

xinglong-zhang:
I believe this is fixed? Can you test with latest main branch? If so, please close this issue. @liu-jing-yi @ZihanL1

PR #226, opened 2025-08-05
fixing issue #223

liu-jing-yi:
fixing problem in extracting the last SCF energy from .log files that contain link jobs


xinglong-zhang:
It looks like there are additional commits/changes in this PR that is not merged into main, such as xtb and xyz tests, can you please open a new PR when needed to merge these commits/changes? @liu-jing-yi

PR #249, opened 2025-08-26
Fix freq def

liu-jing-yi:
fixing the all_structures property of getting all structures for link jobs.


chemsmart/io/gaussian/output.py
```diff
@@ -179,43 +179,40 @@ def all_structures(self):
...
-        if self.job_type == "link" and len(orientations) > 1:
-            orientations, orientations_pbc = (
-                orientations[1:],
-                orientations_pbc[1:],
-            )
-            self.energies = self.energies[1:]
+        if getattr(self, "is_link", False):
+            orientations.pop(0)
+            if self.job_type == "sp":
+                energies = self.energies[2:]
```
xinglong-zhang, line 186: first structure is removed but energies are taken from 3rd one onwards, if there are only two energies, this will give IndexError.
For link sp job, we can simply take last structure/orientation and last energy

chemsmart/io/gaussian/output.py
```diff
@@ -179,43 +179,40 @@ def all_structures(self):
...
-        clean_duplicate_structure(orientations)
-
         # Remove first structure if it's a link job
-        if self.job_type == "link" and len(orientations) > 1:
-            orientations, orientations_pbc = (
-                orientations[1:],
-                orientations_pbc[1:],
-            )
-            self.energies = self.energies[1:]
+        if getattr(self, "is_link", False):
```
xinglong-zhang, line 183: we can use if `self.is_link`, since `is_link` property always has T/F value after initialisation

chemsmart/io/gaussian/output.py
```diff
@@ -179,43 +179,40 @@ def all_structures(self):
...
+            orientations.pop(0)
+            if self.job_type == "sp":
+                energies = self.energies[2:]
+            elif self.job_type == "opt" or self.job_type == "ts":
```
xinglong-zhang, line 187: for `opt` and `ts` jobs, first energy is excluded, but why is this so? and why this is not done for other jobs such as `modred`, `scan` etc?

chemsmart/io/gaussian/output.py
```diff
@@ -179,43 +179,40 @@ def all_structures(self):
...
+                num_structures=num_structures_to_use,
             )
         else:
             # Handle abnormal termination
-            num_structures_to_use = min(
-                len(orientations),
-                len(self.energies),
-                len(self.forces),
-            )
             all_structures = create_molecule_list(
```
xinglong-zhang, line 212: this else condition for creating `all_structures` is the same as if block, so why do we need the conditional?

chemsmart/io/gaussian/output.py
```diff
@@ -234,11 +232,21 @@ def all_structures(self):
                 all_structures[i] for i in self.optimized_steps_indices
             ]
 
-        logger.debug(
-            f"Total number of structures located: {len(all_structures)}"
-        )
+        logger.debug(f"Total number of structures located: {num_structures}")
```
xinglong-zhang, line 235: is `num_structures = len(all_structures) `only used in the logger message? If so we can do

> logger.debug(f"Total number of structures located: {len(all_structures) }")

and remove this assignment `num_structures = len(all_structures) `

chemsmart/io/gaussian/output.py
```diff
@@ -234,11 +232,21 @@ def all_structures(self):
...
 
+    def _get_num_structures_to_use(self):
+        num_structures_to_use = []
+        if self.standard_orientations:
+            num_structures_to_use.append(len(self.standard_orientations))
+        if self.energies:
+            num_structures_to_use.append(len(self.energies))
+        if self.forces:
+            if self.job_type == "link":
+                self.forces.insert(0, None)
```
xinglong-zhang, line 246: instead of inserting force as None, it's better to remove that molecule with (orientation, energy, None force), since all other molecules collected will have a force, except the first one, causing inconsistency. Also note that for link job, the first structure is the same as the first structure after link route.

chemsmart/io/gaussian/output.py
```diff
@@ -645,12 +653,12 @@ def frozen_atoms_masks(self):
 
         return masks
 
-    @cached_property
+    @property
```
xinglong-zhang, line 656: any particular reason to change `cached_property` to `property`? if the calculation is expensive, then we can store as `cached_property`, which only involves calcs once.

chemsmart/utils/io.py
```diff
@@ -19,6 +19,7 @@ def create_molecule_list(
 ):
     """Helper function to create Molecule objects."""
     num_structures = num_structures or len(orientations)
+    print(energies, num_structures)
```
xinglong-zhang, line 22: debugging print statements should be removed.

xinglong-zhang (commented):
Some comments for follow up. @liu-jing-yi

chemsmart/io/gaussian/output.py
liu-jing-yi: Changed.

chemsmart/io/gaussian/output.py
liu-jing-yi: Fixed. Now only the 'else:' block contains the 'num_structures=num_structures_to_use' attribute, which handles the cases where the length of orientation, energies, and forces are unmatched for output log files with abnormal termination.

chemsmart/io/gaussian/output.py
liu-jing-yi: Removed.

chemsmart/io/gaussian/output.py
liu-jing-yi: Thanks for explaining. This has been changed back to 'cached_property'.

chemsmart/utils/io.py
liu-jing-yi: removed.

xinglong-zhang:
your commit [49008c5](https://github.com/liu-jing-yi/chemsmart/commit/49008c546a631496be136513e33c24b811af5627) is not reflected in the file changes in this PR. This PR may have been broken as the repo is made public, can you open a new PR from your fix_freq_def branch? @liu-jing-yi

PR #257, opened 2025-08-28
newest Docs sphinx update to a new branch

ZihanL1:


ZihanL1:
Update newest version to a new branch named **readthedocs** (use Commit 55b199f
 "added citations in README.md" to create). Now it should be ok. 
Update _/build_ folder may lead to some tracking problems like today, so I change the _.gitignore_ file to not push this folder, people who work on readthedocs can run `make html` to see the html in local but should not upload it to GitHub.

PR #274, opened 2025-08-30
renamed yaml files from test to example since people may create their own test yaml project settings

xinglong-zhang:


xinglong-zhang:
Not needed, as examples in README calls test.yaml, users should note that they will need to change the functionals in their own project.yaml if they need other settings

PR #288, opened 2025-09-02
seperate configuration update again

ZihanL1:


ZihanL1:
`exit 0` in _skip tests if only docs changed_ don't skip the whole build test

PR #292, opened 2025-09-03
Link fix

xinglong-zhang:
Fix to obtaining strucutre from link job output + fix to SOMO energy for high spin multiplicity:

the SOMO energy should be the `alpha_occ_eigenvalues[-n] `where n is number of unpaired electtrons of the system.


xinglong-zhang (commented):
Reviewed.

PR #303, opened 2025-09-10
Fix grouper and test, from fix_grouper to main

ZihanL1:


ZihanL1:
This is the latest version of _grouper_, passed all test.

xinglong-zhang:
Thank you, @ZihanL1!

chemsmart/utils/grouper.py
```diff
@@ -159,14 +159,21 @@ def _calculate_rmsd(self, idx_pair: Tuple[int, int]) -> float:
...
+
 
 class RMSDGrouperSharedMemory(MoleculeGrouper):
     """Group molecules based on RMSD using shared memory with minimal locking."""
 
     def __init__(
         self,
         molecules: Iterable[Molecule],
-        threshold: float = 0.5,  # RMSD threshold for grouping
+        threshold: float = 3.5,  # RMSD threshold for grouping
```
xinglong-zhang, line 176: Do we use 0.5 or 3.5 for the `RMSDGrouper`? We should keep the value the same for both `RMSDGrouper` and `RMSDGrouperSharedMemory`.

chemsmart/utils/grouper.py
```diff
@@ -621,15 +628,44 @@ def group(self) -> Tuple[List[List[Molecule]], List[List[int]]]:
 
 class StructureGrouperFactory:
     @staticmethod
-    def create(structures, strategy="rdkit", num_procs=1, **kwargs):
+    def create(
+        structures,
+        strategy="rmsd",
+        num_procs=1,
+        threshold=5.0,
```
xinglong-zhang, line 635: How do we decide to set the threshold value here as 5.0?

xinglong-zhang (commented):
All changes look good thanks @ZihanL1, please look through the comments at your convenience.

chemsmart/utils/grouper.py
ZihanL1: changed to none then use default value from corresponding strategy.

chemsmart/utils/grouper.py
```diff
@@ -632,7 +632,7 @@ def create(
         structures,
         strategy="rmsd",
         num_procs=1,
-        threshold=5.0,
+        threshold=None,
```
xinglong-zhang, line 635: In the individual group type, the value for threshold in the initializer is not None and there is no default value set for the case when threshold=None, e.g., 

> if threshold=None:
>     threshold=0.5

This will cause error when creating the grouper instance:

>         if strategy in groupers:
>             logger.info(f"Using {strategy} grouping strategy.")
>             if strategy in threshold_supported:
>                 if strategy == "rmsd":
>                     return groupers[strategy](
>                         structures,
>                         threshold=threshold,
>                         num_procs=num_procs,
>                         ignore_hydrogens=ignore_hydrogens,
>                         **kwargs,
>                     )

. 

Can you make it consistent that all grouper types use `threshold=None` in the initializer and include an initialization to default value when `threshold=None`, for example:

> if threshold=None:
>     threshold=0.5

xinglong-zhang (changes requested):
Please fix accordingly @ZihanL1

chemsmart/utils/grouper.py
```diff
@@ -266,11 +275,13 @@ class TanimotoSimilarityGrouper(MoleculeGrouper):
...
         self,
         molecules: Iterable[Molecule],
-        threshold: float = 0.9,  # Tanimoto similarity threshold
+        threshold=None,  # Tanimoto similarity threshold
         num_procs: int = 1,
         use_rdkit_fp: bool = True,  # Allows switching between RDKit FP and RDKFingerprint
     ):
         super().__init__(molecules, num_procs)
+        if threshold is None:
+            threshold = 0.9
```
xinglong-zhang, line 284: Is this 0.9 intentional, or should it be 0.5?

xinglong-zhang: Is this already in the docstrings? If not, please add in.

chemsmart/jobs/gaussian/traj.py
```diff
@@ -121,9 +121,13 @@ def __init__(
...
         self.molecules = molecules[-last_num_structures:]
-        self.grouper = StructureGrouperFactory.create(
-            self.molecules, strategy=self.grouping_strategy, **kwargs
-        )
+        if grouping_strategy is not None:
+            self.grouper = StructureGrouperFactory.create(
+                self.molecules, strategy=self.grouping_strategy, **kwargs
+            )
+            self.grouper.group()
+        else:
```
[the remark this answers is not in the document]
xinglong-zhang: added in commit https://github.com/xinglong-zhang/chemsmart/pull/303/commits/62cea3639e8b7f5d6024d7db3f05a54ece5baad4

PR #312, opened 2025-09-20
feat: delete scratch if job is completed successfully

xinglong-zhang:
Jobrunner takes boolean argument `--delete-scratch`, which when set to `True`, will delete the scratch directory if job was run in scratch. Defaults to `False`.


xinglong-zhang (commented):
Thanks @Lewen-WangCC, I made some cleanup and will merge once all tests passed.

PR #313, opened 2025-09-20
updated additional citations

xinglong-zhang:
citations to dependencies of chemsmart package, if used.


xinglong-zhang (commented):
added citations to dependencies

PR #318, opened 2025-09-27
fix spin/nci.pse problem

ZihanL1:


xinglong-zhang (approved):
Thanks @ZihanL1, so `nci` generates the normal `basename.pse `(without _nci suffix), but `spin` generates the correct `basename_spin.pse`?

ZihanL1:
Now nci generate `basename_nci.pse` or `basename_nci_binary.pse `or ` basename_nci_intermediate.pse` and spin generate `basename_spin.pse`. Don't influence the `basename.pse`

xinglong-zhang:
So no basename.pse is generated when nci or spin is run? And if basename.pse exists, it won't be affected, that's what you meant?

ZihanL1:
Yes.

PR #319, opened 2025-09-28
Spin fix

xinglong-zhang:


ZihanL1:
Yes, is a much cleaner way of implementation. I downloaded it and tested, the functions are working well,  except the last removing part need to be revised as when we run other job type, it will delete `'basename.err'` but keep `'base name_jobtype.err'` maybe after you merge this PR.

PR #329, opened 2025-11-11
Data assembler

Huiwen-Tan:


chemsmart/io/gaussian/output.py
```diff
@@ -366,6 +371,32 @@ def align_lengths_to_orientations():
         if self.num_vib_frequencies:
             all_structures[-1] = self._attach_vib_metadata(last_mol)
 
+        # Tag optimized structures
+        try:
```
xinglong-zhang: I am thinking we can store all molecules with energy and forces, even if they are intermediate points and not final optimised structures. These can be useful for MLIP training later on.

PR #334, opened 2025-11-17
Fix scripts

Huiwen-Tan:


chemsmart/io/orca/output.py
```diff
@@ -549,6 +549,25 @@ def multiplicity(self):
...
+            if "Kohn-Sham wavefunction type" in line:
+                wavefunction_type = line.split(" ")[-1].lower()
+                # Determine if the job is restricted or unrestricted
+                if wavefunction_type.startswith("r"):
+                    spin = "restricted"
+                elif wavefunction_type.startswith("u"):
+                    spin = "unrestricted"
+                else:
+                    spin = None
+                return spin
```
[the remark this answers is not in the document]
Huiwen-Tan: yes, should also consider HF, fixed in commit https://github.com/xinglong-zhang/chemsmart/pull/334/commits/0a2beee6cfa9be1341fcdfed372b48bf5dbead61.

chemsmart/io/orca/output.py
```diff
@@ -549,6 +549,29 @@ def multiplicity(self):
                 return int(line.split()[-1])
         return None
 
+    @property
```
xinglong-zhang: can we add in the test for this property in the existing IO test for ORCA output files?

chemsmart/scripts/mulliken.py
```diff
@@ -64,28 +64,29 @@ def entry_point(filename, numbers):
...
-    logger.info("\nMulliken Spin densities:")
-    for hkey, hvalue in mulliken_spins.items():
-        logger.info(f"{hkey:<6}  :  {hvalue:>8.3f}")
-    logger.info("\n")
+    if outputfile.spin == "unrestricted":
+        mulliken_spins = outputfile.mulliken_spin_densities
+        logger.info("\nMulliken Spin densities:")
+        for hkey, hvalue in mulliken_spins.items():
+            logger.info(f"{hkey:<6}  :  {hvalue:>8.3f}")
+        logger.info("\n")
```
xinglong-zhang: We have to check that `outputfile.spin`  is not None first, then check the value of it, right? can we also add an `elif outputfile.spin == "restricted":` and `else: raise Error`?

Huiwen-Tan: I added check for `spin` in commit https://github.com/xinglong-zhang/chemsmart/pull/334/commits/b1465fb5db4a959d47dfe56055c60271ae8236b9, if we add `elif outputfile.spin == "restricted":` might lead to code redundancy.

chemsmart/io/orca/output.py
```diff
@@ -549,6 +549,29 @@ def multiplicity(self):
...
+        the calculation uses restricted (R) or unrestricted (U) spin.
+        """
+        for line in self.contents:
+            if (
+                "Kohn-Sham wavefunction type" in line
+                or "Hartree-Fock type" in line
+            ):
+                wavefunction_type = line.split()[-1].lower()
+                # Determine if the job is restricted or unrestricted
+                if wavefunction_type.startswith("r"):
```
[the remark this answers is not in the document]
Huiwen-Tan: changed in commit https://github.com/xinglong-zhang/chemsmart/commit/b1465fb5db4a959d47dfe56055c60271ae8236b9.

chemsmart/io/orca/output.py
Huiwen-Tan: added in https://github.com/xinglong-zhang/chemsmart/commit/b1465fb5db4a959d47dfe56055c60271ae8236b9.

xinglong-zhang (approved):

PR #352, opened 2025-11-27
bug fix: handle mixed scan/frozen coordinates in Gaussian modred parsing

Lewen-WangCC:
When using a Gaussian log file (mixed scan mission that contain both scan coordinates and freeze coordinates) as input, the program would crash if the Modredundant block contained both scan coordinates (S) and frozen coordinates (F).

The parser previously assumed that if a scan was detected, every line in the Modredundant block was a scan coordinate. This caused an AssertionError when the parser encountered a line containing the F tag (e.g., 29 37 F), as it expected only integer atom indices for scan definitions.

The Fix I have updated the GaussianFileMixin class in mixins.py to handle mixed coordinate definitions robustly:
Improved Job Type Detection: _get_modredundant_conditions now scans all lines in the block. If any scan coordinate (S) is found, the job is identified as a scan; otherwise, it falls back to a frozen coordinate (modred) job.

Robust Parsing: _get_modred_scan_coords: Now explicitly skips lines that do not contain the S tag (ignoring frozen coordinates). get_modred_frozen_coords: Now explicitly skips lines that do not contain the F tag (ignoring scan coordinates).
image


chemsmart/utils/mixins.py
```diff
@@ -351,6 +357,8 @@ def _get_modred_scan_coords(modred_list_of_string):
         coords = []
         # modred = {'num_steps': 10, 'step_size': 0.05, 'coords': [[1, 2], [3, 4]]}
         for raw_line in modred_list_of_string:
+            if "S" not in raw_line and "s" not in raw_line:
```
xinglong-zhang: This `_get_modred_scan_coord() `would ignore frozen coordinates, is this intended?

xinglong-zhang (changes requested):
Two comments for you to think about @Lewen-WangCC

PR #356, opened 2025-11-27
merge main into desc

xinglong-zhang:


xinglong-zhang:
Incorporated in new PR https://github.com/xinglong-zhang/chemsmart/pull/357

Issue #368, opened 2025-11-29
visualization job breaks after incorporation of hybrid visualization job


xinglong-zhang:
Please troubleshoot for why the visualization job breaks in the hybrid branch - I ran the same basic visualisation in main branch and it works, but when i run the same thing in hybrid branch it doesn't (same failure with opening the resulting pymol .pse file)

xinglong-zhang:
Fiexed in PR #375

PR #370, opened 2025-11-29
fix mol runner so that the stdout is written correctly to logfile and not outputfile

xinglong-zhang:


xinglong-zhang:
Fixes issue https://github.com/xinglong-zhang/chemsmart/issues/368

PR #412, opened 2025-12-18
urgent fix to multiple indices for jobs that require only last geometry by default

xinglong-zhang:
Only last structure should be used for creation of input if no index is specified.


xinglong-zhang (commented):
approve

PR #439, opened 2026-01-18
some ctx fix for mol align

xinglong-zhang:


xinglong-zhang:
I have asked copilot to diagnose and work on this and you may follow up @ZihanL1

xinglong-zhang (commented):
looks okay

PR #441, opened 2026-01-18
fix_align_job 1.19

ZihanL1:
1. Add a new _align_ job judgment method to support single-file processing.
2. `-i 1,5-10,-1` can used for single-file alignment, (default _index = :_ for single-file and default _index = -1_ for multiple files now)
3. The current alignment now supports batch processing for over **100** structures and can handle alignment for **500+** molecules.
4. Update the RTD and test.

### DIFF
--- a/tests/test_io_utils.py
+++ b/tests/test_io_utils.py
@@ -267,3 +267,178 @@ class TestLoadMoleculesFromPaths:
         )
         assert len(molecules) > 0
         assert all(mol is not None for mol in molecules)
+
+
+class TestSelectItemsByIndex:
+    """Tests for the select_items_by_index function."""
+
+    def setup_method(self):
+        """Set up test fixtures."""
+        from chemsmart.utils.io import select_items_by_index
+
+        self.select_fn = select_items_by_index
+        self.test_list = ["a", "b", "c", "d", "e", "f", "g", "h", "i", "j"]
+
+    # ---- Tests for None or ":" (return all items) ----
+    def test_none_returns_all(self):
+        """Test that None index_spec returns all items."""
+        result = self.select_fn(self.test_list, None)
+        assert result == self.test_list
+
+    def test_colon_returns_all(self):
+        """Test that ':' index_spec returns all items."""
+        result = self.select_fn(self.test_list, ":")
+        assert result == self.test_list
+
+    # ---- Tests for single index ----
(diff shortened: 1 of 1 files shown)


chemsmart/cli/mol/align.py
```diff
@@ -86,21 +99,124 @@ def align(
...
+                add_index_suffix_for_single=True,
+                check_exists=True,
+            )
+
+            # Now apply user's index selection if specified
+            if index and index != ":":
+
+                def validate_and_add_index(idx, desc=""):
+                    """Helper function to validate index and check for overlaps"""
+                    if idx < 0 or idx >= len(all_molecules):
```
xinglong-zhang: negative indices should also be accepted to be understood as counting from the back, ie, -1 means last structure and -5 means fifth last structure.

chemsmart/cli/mol/align.py
```diff
@@ -86,21 +99,124 @@ def align(
...
+                        raise click.BadParameter(
+                            f"Index overlap detected{desc}."
+                        )
+                    selected_positions.add(idx)
+                    selected_molecules.append(all_molecules[idx])
+
+                selected_molecules = []
+                selected_positions = set()  # Track actual 0-based positions
+
+                if "-" in index:
```
xinglong-zhang: would be invalidated by previous 
`if idx < 0 `
?

chemsmart/cli/mol/mol.py
```diff
@@ -374,24 +374,36 @@ def mol(
...
+                filenames = filenames[0]
+                molecules = Molecule.from_filepath(
+                    filepath=filenames, index=":", return_list=True
+                )
+                assert (
+                    molecules is not None
+                ), f"Could not obtain molecule from {filenames}!"
+                logger.debug(f"Obtained molecule {molecules} from {filenames}")
+            else:
+                raise ValueError(
```
xinglong-zhang: I thought our redesigned codes will allow multiple files for all jobs? I may be wrong though.

chemsmart/utils/io.py
```diff
@@ -436,6 +436,56 @@ def load_molecules_from_paths(
     return loaded
 
 
+def select_items_by_index(
```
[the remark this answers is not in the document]
ZihanL1: Added in the end of _test_io_utils.py_ and passed. Please have a check

PR #447, opened 2026-01-22
Data assembler

xinglong-zhang:


xinglong-zhang:
@Huiwen-Tan, please follow up on copilot comments.

PR #490, opened 2026-03-05
feat(molecules): add robust InChIKey property via Open Babel

Lewen-WangCC:
Implement the `inchikey` property in the `Molecule` class.

This utilizes `pybel.readstring` to parse an in-memory XYZ string, leveraging Open Babel's robust topological bond perception. This bypasses the native distance-based bond order guessing which often causes sanitization failures in RDKit, especially for organometallic complexes and transition states with unusual valences.

### DIFF
--- a/tests/test_structures.py
+++ b/tests/test_structures.py
@@ -1945,7 +1945,7 @@ class TestInChIKey:
             mol = mol[-1]
         return mol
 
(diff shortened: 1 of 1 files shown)


xinglong-zhang (changes requested):
Please add in appropriate tests for this. Remember to also test edge cases.

PR #497, opened 2026-03-10
update RTD for gitbash installation

ZihanL1:
Tested and recorded the installation on Windows using Gitbash (pip install chemsmart), and updated the RTD for it. Now we recommend users to use WSL or Ubuntu for chemsmart but also can use Gitbash to install chemsmart on windows.

### DIFF
--- a/chemsmart/cli/gaussian/modred.py
+++ b/chemsmart/cli/gaussian/modred.py
@@ -48,6 +48,11 @@ def modred(
         project_settings, jobtype, coordinates, step_size, num_steps
     )
 
+    # Merge gaussian-level CLI overrides (e.g. -c/--charge, -m/--multiplicity).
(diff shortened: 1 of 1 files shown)


docs/source/installation-windows-gitbash.rst
```diff
@@ -58,14 +60,8 @@ This guide covers installing Chemsmart on Windows using Git Bash.
  Make Installation
 *******************
 
-#. Install the package and dependencies:
+#. Install the package and dependencies via pip command:
```
xinglong-zhang: so `make install` and `make install-dev` do not work anymore? Earlier version installs using these and usually when more dependencies are needed after an initial installation, one can call either `make install` or `make install-dev` to update those dependencies. If `make install` and `make install-dev` do not work, then I will merge this and keep it using `pip install chemsmart` for now.

PR #503, opened 2026-03-14
fix to support condashell and gitbash installation and update RTD

ZihanL1:
Now it support condashell installation and gitbash installation, with record.


xinglong-zhang (changes requested):
just one comment before i merge and we can come back to this later.

PR #512, opened 2026-03-17
feat(converter): add CDX/CDXML support with reference output files & add IterateMoleculePool class to reduce memory burden under potential extreme conditions 

Lewen-WangCC:
- FileConverter now handles cdxml/cdx formats via CDXFile.molecules
- Multi-molecule CDXML splits into basename_1.ext, basename_2.ext, ...
- Add 10 reference output files under tests/data/.../chemdraw/expected/
- Add conftest fixtures for expected files (xyz + com)
- Update converter tests to compare positions/formula against reference files


Lewen-WangCC:
Added IterateMoleculePool as a shared molecule pool for ITERATE, to reduce the memory burden under potential extreme conditions. Details can be seen in the commit @xinglong-zhang

PR #555, opened 2026-05-07
Update GROMACS job and runner structure

Ayson-HAN:
This PR updates the initial GROMACS job and runner structure based on the shared `gromacs` branch.

Current scope:
- keeps `GromacsJob` inherited from the base `Job` class
- keeps `GromacsEMJob` inherited from `GromacsJob`
- adds GROMACS-specific file attributes, including `mdp_file`, `structure_file`, `top_file`, `tpr_file`, `itp_files`, and `index_file`
- updates `GromacsJobRunner` to inherit from the base `JobRunner`
- uses the existing runner lifecycle instead of defining a separate run flow
- adds pre-run validation for required GROMACS input files
- adds initial TPR assembly logic using `gmx grompp`
- updates `mdrun` command generation to use the actual TPR file prefix
- adds smoke tests for job type matching, file attributes, TPR path handling, runner command generation, and input validation

Notes:
- this is still an initial structural implementation
- the first supported path is user-provided/prepared GROMACS input files
- automatic force-field generation is not implemented yet
- settings/YAML integration, writer logic, executable settings, and full end-to-end EM execution remain next steps

### DIFF
--- a/chemsmart/jobs/gromacs/job.py
+++ b/chemsmart/jobs/gromacs/job.py
@@ -29,7 +29,12 @@ class GromacsJob(Job):
         self.mdp_file = Path(mdp_file) if mdp_file else None
         self.structure_file = Path(structure_file) if structure_file else None
         self.top_file = Path(top_file) if top_file else None
-        self.tpr_file = Path(tpr_file) if tpr_file else Path(self.folder) / f"{self.label}.tpr"
+        self._use_default_tpr_file = tpr_file is None
+        self.tpr_file = (
+            Path(tpr_file)
+            if tpr_file is not None
+            else Path(self.folder) / f"{self.label}.tpr"
+)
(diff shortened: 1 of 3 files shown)


xinglong-zhang (approved):

chemsmart/jobs/gromacs/job.py
```diff
@@ -41,6 +46,27 @@ def has_tpr(self):
 
     def has_topology(self):
         return self.top_file is not None and self.top_file.exists()
+    def has_required_prepared_inputs(self):
+        """
+        Check whether the job has the minimum user-provided files
+        required to assemble a TPR file.
+        """
+        return (
+                self.mdp_file is not None
```
liu-jing-yi, line 55: will it be better to add path check such as `Path.exists()`?

PR #560, opened 2026-05-13
change server.yaml in test folder for consistency

ZihanL1:


xinglong-zhang (approved):

xinglong-zhang:
@ZihanL1, this change  will also cause errors in gaussian input files that are written. If update, need to update those expected Gaussian written input files too. (+ ORCA too)

ZihanL1:
image

tests/test_groupers.py
```diff
@@ -837,19 +837,18 @@ def test_count_groups_matches_complete_linkage_iteration_order(
         self, methanol_molecules
     ):
         grouper = TorsionFingerprintGrouper(
-            methanol_molecules[:4],
+            methanol_molecules[:3],
```
xinglong-zhang, line 840: what happened here? seems unrelated to the change in MEM_GB value in server.yaml?

xinglong-zhang (changes requested):
@ZihanL1 , what cause the need to change the test_groupers.py?

PR #562, opened 2026-05-14
update the NCIPLOT tutorial

ZihanL1:
image
I think the input for NCIPLOT makes the difference so I add one warning window in the NCIPLOT tutorial. Also add the doctoring in the codes.


xinglong-zhang (changes requested):
Looks good thanks, @ZihanL1. Can you add in the bash commands for running nciplot to generate promolecular vs wavefunction density nciplots? `chemsmart sub nciplot -f file.log` vs `chemsmart sub nciplot -f file_nci.wfn`.

ZihanL1:
image
Added. This is the current warning window.

xinglong-zhang:
> image
> 
> Added. This is the current warning window.

Can you add in additional explanation of the type of cube files generated, since these two commands do not directly give you the plots.

Please update the workflow: run gaussian nci to obtain wfn, run nciplot to obtain cube files then run mol command to plot.

ZihanL1:
image
image
add workflow

xinglong-zhang:
> image image add workflow

Looks great, thanks!

docs/source/_static/nciplot-workflow.mmd
xinglong-zhang, line 1: this file is empty and should be removed?

xinglong-zhang (changes requested):
There are two empty files, should they be removed?

docs/source/_static/nciplot-workflow.mmd
ZihanL1: yes, I removed them

xinglong-zhang (approved):

xinglong-zhang:
Thanks, @ZihanL1, merging now.

PR #563, opened 2026-05-16
feat(pka): add end-to-end pKa workflows for Gaussian/ORCA

liu-jing-yi:
1. Adds full pKa submission and analysis support across Gaussian and ORCA, including a backend-independent run pka analysis path.
2. Introduces new pKa job/settings implementations, thermodynamic cycle handling (proton exchange/direct), and improved batch submission behavior with per-entry CLI reconstruction.
3. Expands CDXML/ChemDraw proton detection and pKa file/output parsing utilities.
4. Updates scheduler/submitter integration for batch workflows and adds broad test coverage (tests/test_pka.py + related fixtures).
5. Adds pKa documentation pages for usage and workflow guidance.


xinglong-zhang:
Please fix the failed tests.

liu-jing-yi:
> Please fix the failed tests.

CI errors fixed. Please have a review.

chemsmart/cli/pka.py
```diff
@@ -0,0 +1,886 @@
...
+
+from chemsmart.utils.cli import MyCommand, MyGroup
+from chemsmart.utils.io import (
+    detect_program_type_from_files,
+    get_program_type_from_file,
+)
+
+logger = logging.getLogger(__name__)
+
+_EXT_MAP = {
```
xinglong-zhang: Do we already use this somewhere in the main branch? If so we can import this; if not, this map should be somewhere higher level, as I imagine this will be used elsewhere outside pka too.

chemsmart/cli/pka.py
```diff
@@ -0,0 +1,886 @@
...
+              multiple molecules are detected in CDX/CDXML.
+
+    Raises:
+        click.UsageError: If required inputs are missing or inconsistent with
+            the file type.
+    """
+    from chemsmart.io.file import PKaCDXFile
+
+    try:
+        return PKaCDXFile.resolve_proton_index(
```
xinglong-zhang: this code shows that this resolve_proton_index works only for PKaCDXFile, then why is there a need to define this function?

chemsmart/cli/pka.py
```diff
@@ -0,0 +1,886 @@
...
+    if shared["reference_multiplicity"] is None:
+        missing.append("-rm/--reference-multiplicity")
+    if missing:
+        raise click.UsageError(
+            "When --reference is provided, the following options are "
+            "required: " + ", ".join(missing)
+        )
+
+
+def run_pka_from_output_table(
```
xinglong-zhang: Looks cool, wondering if this output_table can be a `class pKaOutputTable` object, with `pKaOutputTable.run_pka()` method? This allows creation of a pKaOutputTable object that may be later on passed as constructor in other classes for interaction.

xinglong-zhang: If every time this (and other related functions are called, it requires a pKa output table as an argument, then perhaps it's best to construct as a class?

chemsmart/cli/pka.py
```diff
@@ -0,0 +1,886 @@
...
+    if program == "gaussian":
+        from chemsmart.io.gaussian.output import Gaussian16pKaOutput
+
+        return Gaussian16pKaOutput
+    from chemsmart.io.orca.output import ORCApKaOutput
+
+    return ORCApKaOutput
+
+
+@click.group(name="pka", cls=MyGroup)
```
xinglong-zhang, line 1191: no need pressure options? looks like these are options available in themochemistry? also we may decompose thermochemistry options into more specialised options that can be reused, eg.,

def thermochemistry_temp_pressure_conc_options:
    etc

def thermochemistry_cutoff_options:
    etc

then can do mix and match. Just an idea for writing fewer lines of duplicated codes.

chemsmart/cli/pka.py
```diff
@@ -0,0 +1,886 @@
...
+                program = get_program_type_from_file(val)
+                break
+        if program == "auto" or program == "unknown":
+            raise click.UsageError(
+                "Could not auto-detect QC program from output table.  "
+                "Use -p gaussian or -p orca."
+            )
+        logger.info(f"Auto-detected program: {program}")
+
+    run_pka_from_output_table(
```
xinglong-zhang: if in OOP, then can create output_table = pKaOutputTable, then you have output_table.run_pka(...)

chemsmart/cli/gaussian/pka.py
```diff
@@ -0,0 +1,403 @@
...
+    """
+    shared = ctx.obj["pka_shared"]
+    filename = ctx.obj.get("filename")
+    proton_index = ctx.obj.get("pka_proton_index")
+    color_code = ctx.obj.get("pka_color_code")
+    jobrunner = ctx.obj["jobrunner"]
+
+    # ── resolve proton index (CDXML auto-detect) ──
+    proton_index, pka_molecules = resolve_proton_index(
+        filename, proton_index, color_code
```
xinglong-zhang, line 159: the resolve_proton_index, as coded up, will raise error for any filename except cdxml/cdx file, is this intended behaviour?

xinglong-zhang: If so, then it's better to have something like this, using OOP (see earlier comments for this feature implementation):

> try:
>     pka_file = PKaCDXFile(filename)
> except SomeError:
>     raise SomeError(f'filename {filename} not supported, requires cdxml)

chemsmart/io/molecules/structure.py
```diff
@@ -2564,6 +2683,136 @@ def pbc_conditions(self):
             return None
 
 
+class SDFFile(FileMixin):
```
xinglong-zhang: This class was already moved to somewhere else and doesn't exist in main branch here. Did you perhaps reintroduce these while fixing conflicts (fixing conflicts should follow main branch if the conflicts do not come from the feature you implement)?

chemsmart/jobs/gaussian/pka.py
```diff
@@ -0,0 +1,890 @@
...
+                settings.reference_pair_molecules()
+            )
+        return cls._shared_reference_molecule_cache[cache_key]
+
+    @classmethod
+    def from_molecules(
+        cls,
+        pka_molecules,
+        shared,
+        project_settings,
```
xinglong-zhang: seems to have too many things you are dealing with, the project_settings and job_settings are only merged in CLI. the Job class only requires the molecule and settings.

chemsmart/jobs/gaussian/pka.py
```diff
@@ -0,0 +1,890 @@
...
+            keywords: Gaussian keywords to merge into opt settings.
+            jobrunner: JobRunner instance for execution.
+            filename: Source filename used to derive fragment labels.
+            skip_completed: Skip execution for completed jobs.
+            **kwargs: Additional job arguments forwarded to GaussianpKaJob.
+
+        Returns:
+            list[GaussianpKaJob]: List of created pKa jobs (no batch wrapper).
+        """
+        opt_settings = project_settings.opt_settings()
```
xinglong-zhang: should not be introduced here. the final setting should be assemble at CLI, as per all other jobs. This implementation may introduce confusing to later-coming developers.

chemsmart/jobs/gaussian/pka.py
```diff
@@ -0,0 +1,890 @@
...
+from chemsmart.jobs.gaussian.job import GaussianJob
+from chemsmart.jobs.gaussian.opt import GaussianOptJob
+from chemsmart.jobs.gaussian.settings import GaussianpKaJobSettings
+from chemsmart.jobs.gaussian.singlepoint import GaussianSinglePointJob
+from chemsmart.jobs.runner import decide_phase_transition, run_phase_jobs
+
+logger = logging.getLogger(__name__)
+
+
+def build_gaussian_pka_settings(
```
xinglong-zhang: settings should be returned in jobs.gaussian.settings.py? also from the codes, it seems that this should be done in the CLI and the finally built setting be supplied to pKaJob.

chemsmart/jobs/gaussian/pka.py
```diff
@@ -0,0 +1,890 @@
...
+        out = job._output()
+        if out is None:
+            return (
+                f"[{role}] Optimisation produced no output – "
+                f"cannot validate frequencies for {job.label}."
+            )
+        if not out.normal_termination:
+            return None  # abnormal termination handled separately
+
+        freqs = getattr(out, "vibrational_frequencies", None)
```
xinglong-zhang: don't use getattr! output file should have already vibrational_frequencies initialised when the object is instantiated.

chemsmart/jobs/gaussian/pka.py
```diff
@@ -0,0 +1,890 @@
...
+            self.ref_acid_sp_job,
+            self.ref_conjugate_base_sp_job,
+        ]
+
+    # ------------------------------------------------------------------
+    # Imaginary frequency validation
+    # ------------------------------------------------------------------
+
+    @staticmethod
+    def _validate_imaginary_frequencies(job, role):
```
xinglong-zhang: instead of having this here, perhaps we can have a out.validate_imaginary_frequencies() method, as this may be used elsewhere too.

chemsmart/jobs/gaussian/settings.py
```diff
@@ -952,6 +952,929 @@ def _check_solvent(self, solvent_model):
...
+            functional="B3LYP",
+            basis="6-311+G(d,p)",
+            solvent_model="SMD",
+            solvent_id="water"
+        )
+    """
+
+    # Default absolute free energy of H+ in water (kcal/mol)
+    # From Tissandier et al., J Phys Chem A 1998, 102:7787
+    DEFAULT_DELTA_G_PROTON = -265.9
```
xinglong-zhang, line 1055: this value is in water and will it be used wrongly if value is needed in other solvents?

chemsmart/jobs/gaussian/settings.py
```diff
@@ -952,6 +952,929 @@ def _check_solvent(self, solvent_model):
...
+        reference_conjugate_base_multiplicity=None,
+        delta_G_proton=None,
+        solvent_model="SMD",
+        solvent_id="water",
+        conjugate_base_charge=None,
+        conjugate_base_multiplicity=None,
+        # Thermochemistry settings for pKa calculation
+        temperature=298.15,
+        concentration=1.0,
+        pressure=1.0,
```
xinglong-zhang, line 1075: pressure option was missing from CLI.

xinglong-zhang (changes requested):
Thanks @liu-jing-yi, this is a nice work and there are many lines of codes. One major comment is that you may try to structure some functionalities as an object, so that the object may be passed to other classes, if needed. See detailed comments in various places.

chemsmart/cli/pka.py
liu-jing-yi: Thank you for pointing this out. `PROGRAM_INFO` in `io.py` does the same thing. The code has been refactor to use `PROGRAM_INFO`.

chemsmart/cli/pka.py
liu-jing-yi: Refactored to PKaOutputTable class.

chemsmart/cli/pka.py
liu-jing-yi: Refactored. `thermochemistry` decorators are introduced and reused in `pka.py`.

chemsmart/cli/pka.py
liu-jing-yi: Refactored.

chemsmart/io/molecules/structure.py
liu-jing-yi: Duplicate codes were removed.

chemsmart/jobs/gaussian/pka.py
liu-jing-yi: Thanks for pointing this out! `build_gaussian_pka_settings` is now handled in `cli.gaussian.pka.py`.

chemsmart/jobs/gaussian/pka.py
liu-jing-yi: Refactored. The job class now only takes a molecule and a finished settings object.

chemsmart/jobs/gaussian/pka.py
liu-jing-yi: Refactored. The job class now only takes a molecule and a finished settings object.

chemsmart/jobs/gaussian/pka.py
liu-jing-yi: Refactored. Now `validate_imaginary_frequencies` is a function in `mixin.py`.

chemsmart/jobs/gaussian/pka.py
liu-jing-yi: Changed.

chemsmart/jobs/gaussian/settings.py
liu-jing-yi: Added.

chemsmart/jobs/gaussian/settings.py
liu-jing-yi: Thanks for pointing this out. The code has been refactored so that the user-specified value will override this default value. The RTD is also updated.

chemsmart/cli/gaussian/pka.py
liu-jing-yi: Refactored.

chemsmart/cli/pka.py
liu-jing-yi: Thanks for pointing this out. The resolve_proton_index has been refactored to handle auto-detection and cli overriding of multiple input format.

chemsmart/cli/gaussian/pka.py
```diff
@@ -0,0 +1,466 @@
...
+@click.pass_context
+def submit(ctx, skip_completed, **kwargs):
+    """Submit a single-molecule Gaussian pKa calculation.
+
+    Builds Gaussian pKa settings from CLI/project context, validates required
+    inputs (charge/multiplicity), and returns the created job(s). If the input
+    is a multi-fragment CDXML, this will expand to one job per molecule.
+
+    Example:
+        chemsmart run gaussian <gaussian_options> pka <pka_options> submit
```
xinglong-zhang: Is this `chemsmart run` or should it be `chemsmart sub`?

chemsmart/cli/gaussian/pka.py
xinglong-zhang: I did not see the refactor. if resolve_proton_index() method only need to be applied to PKaCDXFile object, then it should be a class method.

chemsmart/cli/gaussian/pka.py
```diff
@@ -0,0 +1,466 @@
...
+    input_table_path = ctx.obj.get("filename")
+    if not input_table_path:
+        raise click.UsageError(
+            "Batch mode requires the parent Gaussian -f/--filename to specify the table file path."
+        )
+
+    from chemsmart.io.molecules.structure import Molecule
+    from chemsmart.jobs.gaussian.pka import GaussianpKaJob
+    from chemsmart.utils.utils import (
+        parse_pka_table,
```
xinglong-zhang: again, can these two functions be refactored as class methods of pKaTable object?

chemsmart/cli/gaussian/pka.py
```diff
@@ -0,0 +1,466 @@
...
+    )
+
+    try:
+        entries = parse_pka_table(input_table_path)
+        validate_pka_table_entries(entries, check_file_exists=True)
+    except (FileNotFoundError, ValueError) as e:
+        raise click.UsageError(str(e))
+
+    if shared["scheme"] == "proton exchange":
+        from chemsmart.cli.pka import resolve_reference_proton
```
xinglong-zhang: I see this `resolve_reference_proton` needs to be applied to different types of files? or is it only to cdxml? If so then again can be class method of PKaCDXFile object as suggested previously

chemsmart/cli/thermochemistry/thermochemistry.py
```diff
@@ -21,80 +20,107 @@
...
-    Common click options for Thermochemistry.
-    """
-
-    @click.option(
+def thermochemistry_cutoff_options(
+    f,
+    entropy_grimme_default=None,
+    entropy_truhlar_default=None,
+    enthalpy_default=None,
+    include_truhlar=True,
```
xinglong-zhang: should it be include_truhlar, or as originally, cutoff-entropy-truhlar? it's not simply including a correction but a completely different scheme?

chemsmart/cli/thermochemistry/thermochemistry.py
```diff
@@ -21,80 +20,107 @@
...
-    @click.option(
+    )(f)
+    return f
+
+
+def click_thermochemistry_options(f):
+    """
+    Common click options for Thermochemistry.
+    """
+    f = thermochemistry_temp_pressure_conc_options(f)
```
xinglong-zhang, line 119: oh you can use this directly like this?

chemsmart/cli/thermochemistry/thermochemistry.py
```diff
@@ -103,36 +129,31 @@ def click_thermochemistry_options(f):
...
         "-O",
         "--overwrite",
         is_flag=True,
         default=False,
         show_default=True,
         help="Overwrite existing output files if they already exist.",
-    )
-    @click.option(
+    )(f)
+    return click.option(
```
xinglong-zhang: is this correct?

chemsmart/cli/pka.py
```diff
@@ -0,0 +1,1018 @@
...
+        return Gaussian16pKaOutput
+    elif program == "orca":
+        from chemsmart.io.orca.output import ORCApKaOutput
+
+        return ORCApKaOutput
+    else:
+        raise ValueError(f"Unsupported program: {program}")
+
+
+def echo_pka_output_table_results(
```
xinglong-zhang: class method for pKaTable object?

chemsmart/cli/run.py
```diff
@@ -121,9 +124,26 @@ def process_pipeline(ctx, *args, **kwargs):
                 num_gpus=jobrunner.num_gpus,
                 mem_gb=jobrunner.mem_gb,
             )
-            # Attach jobrunner to job and run the job with the jobrunner
-            single_job.jobrunner = job_specific_runner
-            single_job.run()
+
+        logger.info("Running jobs with concurrent submitters")
```
xinglong-zhang: is this batch submission? have they been tested and the codes optimised?

chemsmart/io/file.py
```diff
@@ -481,52 +505,805 @@ def _preprocess_cdxml_remove_multi_attachments(filepath):
...
+    * providing structured atom and proton information that can be consumed
+      by higher-level pKa logic (for example, to build :class:`PKaMolecule`
+      instances or similar pKa-aware molecule objects).
+
+    The class does **not** perform the pKa calculation itself; instead it
+    focuses on converting CDXML drawings into a representation where
+    ionizable sites and their associated protons can be identified reliably.
+    """
+
+    @classmethod
```
xinglong-zhang: no need @classmethod decorator, since you are not returning the class itself from it

chemsmart/io/file.py
```diff
@@ -481,52 +505,805 @@ def _preprocess_cdxml_remove_multi_attachments(filepath):
...
+            raise ValueError(
+                "-cc/--color-code can only be used with .cdx/.cdxml files."
+            )
+
+        raise ValueError(
+            "-pi/--proton-index is required when launching new pKa "
+            "calculations (or use a .cdxml file with a coloured proton)."
+        )
+
+    @classmethod
```
xinglong-zhang: no need decorator

chemsmart/jobs/gaussian/pka.py
```diff
@@ -0,0 +1,729 @@
...
+        settings (GaussianpKaJobSettings): pKa calculation configuration.
+        label (str): Base job identifier used for file naming.
+        jobrunner (JobRunner): Execution backend that runs the jobs.
+        skip_completed (bool): If True, completed jobs are not rerun.
+    """
+
+    TYPE = "g16pka"
+    _shared_reference_molecule_cache = {}
+
+    @classmethod
```
xinglong-zhang: no need this decorator, please also check elsewhere where i have not commented.

chemsmart/utils/constants.py
```diff
@@ -28,6 +28,10 @@
 hartree_to_joules = 4.35974434e-18  # 1 Hartree = 4.35974434 × 10^-18 Joules
 cal_to_joules = 4.184  # 1 Calorie = 4.184 Joules
 
+# Direct conversion factor: 1 Hartree = 627.5094740631 kcal/mol
```
xinglong-zhang: is this value good enough? different tables might give different conversion factors and the best might be derive this from the constants we have from ase.units.

chemsmart/utils/io.py
```diff
@@ -57,7 +57,7 @@
             "Your ORCA version",
             "ORCA versions",
         ],
-        "suffixes": [".out"],
+        "suffixes": [".out", ".log"],
```
xinglong-zhang, line 61: orca also uses .log extension? if this has been modified, need to add in appropriate tests.

chemsmart/utils/mixins.py
```diff
@@ -392,6 +392,32 @@ def beta_fmo_gap(self):
...
+            return None
+
+        imaginary = [f for f in self.vibrational_frequencies if f < 0.0]
+        if not imaginary:
+            return None
+
+        display_label = label or self.filename
+        role_prefix = f"[{role}] " if role else ""
+        return (
+            f"{role_prefix}Imaginary frequency check FAILED for "
```
xinglong-zhang: does this method only return a message? I thought it would evaluate to True if it contains imaginary frequency and False otherwise. Also, what about TSs with expected imaginary frequency?

chemsmart/utils/utils.py
```diff
@@ -2021,3 +2021,1205 @@ def token_score(token):
...
 
     return " ".join(new_tokens)
+
+
+# ---------------------------------------------------------------------------
+# pKa Table Parsing Utilities
+# ---------------------------------------------------------------------------
+
+
+class TabularDataset:
```
xinglong-zhang: should this class and the class following this be in utils.py or io.py?

chemsmart/utils/utils.py
```diff
@@ -2021,3 +2021,1205 @@ def token_score(token):
...
+            except (TypeError, ValueError):
+                errors.append(
+                    f"Invalid multiplicity: {multiplicity!r}{row_info}"
+                )
+
+        if errors:
+            raise ValueError("; ".join(errors))
+
+
+def parse_pka_table(
```
xinglong-zhang: can be a class method of PKaTableEntry class.

xinglong-zhang: note my comment `a class method` does not mean `@classmethod` !

chemsmart/utils/utils.py
```diff
@@ -2021,3 +2021,1205 @@ def token_score(token):
...
+                except (TypeError, ValueError):
+                    errors.append(
+                        f"Invalid pka_ref: {self.pka_ref!r}{row_info}"
+                    )
+
+        if errors:
+            raise ValueError("; ".join(errors))
+
+
+def parse_pka_output_table(table_path: str, delimiter: str = None) -> list:
```
xinglong-zhang: can these be methods of PkaTable object for better encapsulation? and function below

xinglong-zhang (changes requested):
Thanks, @liu-jing-yi. We are almost there, we can discuss tomorrow afternoon after you have reviewed my comments (and make changes where needed).

chemsmart/utils/utils.py
liu-jing-yi: Personally I prefer to put it in `utils.py`. `io.py` manipulates the `molecule` object, and `utils.py` contains generally utility functions for CHEMSMART. This class is designed to read table-format of job settings attributes CHEMSMART, hence it's better to put it in `utils.py.`

chemsmart/io/file.py
liu-jing-yi: Removed.

chemsmart/jobs/gaussian/pka.py
liu-jing-yi: Thanks for pointing this out! I have removed them.

chemsmart/utils/constants.py
liu-jing-yi: Updated with `ase` constants.

chemsmart/utils/io.py
liu-jing-yi: By default ORCA use .out format but also accepts .log if user specify. Supporting both extension will make our code more flexible  and robust. Corresponding test has been added for this.

chemsmart/utils/mixins.py
liu-jing-yi: Updated. Now this method will perform validation for opt and ts jobs and return a dictionary containing the validation results.

chemsmart/utils/utils.py
liu-jing-yi: Refactored.

chemsmart/utils/utils.py
liu-jing-yi: Refactored.

chemsmart/cli/thermochemistry/thermochemistry.py
liu-jing-yi: Thanks for pointing this out. This is the original cutoff-entropy-truhlar, which follows the original design of thermochemistry module. The code has been updated accordingly.

chemsmart/cli/thermochemistry/thermochemistry.py
liu-jing-yi: Fixed.

chemsmart/cli/run.py
liu-jing-yi: Thanks for pointing this out. This is for batch processing and has been removed, as the `pka_only` branch is solely for PKa calculations. This function is retained in the `batchjob` branch.

chemsmart/cli/pka.py
liu-jing-yi: Refactored.

chemsmart/cli/thermochemistry/thermochemistry.py
liu-jing-yi: Yes. This internally applies `thermochemistry_temp_pressure_conc_options` decorator to that command.

chemsmart/cli/gaussian/pka.py
liu-jing-yi: Should be chemsmart sub. Updated.

chemsmart/cli/gaussian/pka.py
liu-jing-yi: Thanks for pointing this out. There is a duplicate implementation with` resolve_proton_index` in PKaCDXFile. The duplicate codes have been removed hence only the class method in PKaCDXFile is retained.

chemsmart/cli/gaussian/pka.py
liu-jing-yi: Refactored.

xinglong-zhang:
is the refactor for ORCA also done? I may look at it later when i have more time.. @liu-jing-yi

liu-jing-yi:
> is the refactor for ORCA also done? I may look at it later when i have more time.. @liu-jing-yi

I am still testing. I will request a review from you when it's ready.

liu-jing-yi:
### CHEMSMART PKa Implementation (updated on Jun. 12, 2026)

Refactors the pKa workflow: job submission stays on Gaussian/ORCA (pka submit / batch); analysis is backend-independent via chemsmart run pka analyze / batch-analyze (mixed .log/.out supported).

1. Highlights

- Proton exchange (default) and direct (-s direct) thermodynamic cycles with dual-level gas-phase freq + solvent SP
- CSV batch tables and multi-fragment ChemDraw CDXML (pka batch → one job per fragment, coloured-proton auto-detection)
- CDXML rows in CSV may omit proton_index; charge/multiplicity inferred from structures when not set on the CLI
- Cluster scripts reconstruct per row/fragment with explicit --proton-index, --index, --label, charge, and multiplicity

2. Fixes

- Duplicate CDXML batch submissions
- Shared output filenames across fragments
- CSV table input incorrectly passed to molecule loading

3.  Example output (these PKa values are not physically meaningful, but are derived from a test reference acid)
```
==============================================================================
Batch pKa Results (Dual-level Proton Exchange)
==============================================================================
Temperature: 278.15 K
Pressure: 1.0 atm
basename                              pKa   ΔG_soln (kcal/mol)
------------------------------------------------------------------------------
pka_scale_frag1                     -8.17             -23.8913
pka_scale_frag4                     -1.24             -15.0654
pka_scale_frag3                      9.22              -1.7521
pka_scale_frag5                     -2.68             -16.8958
==============================================================================
```

xinglong-zhang:
> ### CHEMSMART PKa Implementation (updated on Jun. 12, 2026)
> Refactors the pKa workflow: job submission stays on Gaussian/ORCA (pka submit / batch); analysis is backend-independent via chemsmart run pka analyze / batch-analyze (mixed .log/.out supported).
> 
> 1. Highlights
> 
> * Proton exchange (default) and direct (-s direct) thermodynamic cycles with dual-level gas-phase freq + solvent SP
> * CSV batch tables and multi-fragment ChemDraw CDXML (pka batch → one job per fragment, coloured-proton auto-detection)
> * CDXML rows in CSV may omit proton_index; charge/multiplicity inferred from structures when not set on the CLI
> * Cluster scripts reconstruct per row/fragment with explicit --proton-index, --index, --label, charge, and multiplicity
> 
> 2. Fixes
> 
> * Duplicate CDXML batch submissions
> * Shared output filenames across fragments
> * CSV table input incorrectly passed to molecule loading
> 
> 3. Example output (these PKa values are not physically meaningful, but are derived from a test reference acid)
> 
> ```
> ==============================================================================
> Batch pKa Results (Dual-level Proton Exchange)
> ==============================================================================
> Temperature: 278.15 K
> Pressure: 1.0 atm
> basename                              pKa   ΔG_soln (kcal/mol)
> ------------------------------------------------------------------------------
> pka_scale_frag1                     -8.17             -23.8913
> pka_scale_frag4                     -1.24             -15.0654
> pka_scale_frag3                      9.22              -1.7521
> pka_scale_frag5                     -2.68             -16.8958
> ==============================================================================
> ```

@copilot, has these features been well-documented in the RTDs? If not, please update RTDs accordingly.

chemsmart/cli/gaussian/pka.py
```diff
@@ -0,0 +1,529 @@
...
+    if is_pka_cdxml_input(input_table_path):
+        return batch_pka_jobs_from_cdxml(
+            ctx,
+            skip_completed,
+            _create_pka_jobs_from_molecules,
+            lambda ctx, **invoke_kwargs: ctx.invoke(submit, **invoke_kwargs),
+            **kwargs,
+        )
+
+    from chemsmart.jobs.gaussian.pka import GaussianpKaJob
```
xinglong-zhang: these imports appear multiple times on this file, maybe just import once at the start of the file.

chemsmart/cli/gaussian/pka.py
```diff
@@ -0,0 +1,529 @@
...
+    job_settings = ctx.obj["job_settings"]
+    keywords = ctx.obj["keywords"]
+    jobrunner = ctx.obj["jobrunner"]
+    filename = ctx.obj.get("filename", "")
+
+    opt_settings = project_settings.opt_settings()
+    if job_settings:
+        opt_settings = opt_settings.merge(job_settings, keywords=keywords)
+
+    base_name = os.path.splitext(os.path.basename(filename))[0] or "pka"
```
xinglong-zhang: use basename for consistency

chemsmart/cli/pka.py
```diff
@@ -0,0 +1,1404 @@
...
+)
+from chemsmart.io.file import PKaCDXFile
+from chemsmart.utils.cli import MyCommand, MyGroup
+from chemsmart.utils.constants import HARTREE_TO_KCAL_MOL, energy_conversion
+from chemsmart.utils.io import get_program_type_from_file
+
+logger = logging.getLogger(__name__)
+
+
+# ── helpers moved from cli/pka_helpers.py ──────────────────────────────
```
xinglong-zhang: delete these comments

chemsmart/cli/orca/pka.py
```diff
@@ -0,0 +1,526 @@
...
+    logger.info(f"Created {len(jobs)} ORCA pKa jobs from table")
+    return jobs
+
+
+# ═══════════════════════════════════════════════════════════════════════
+# Private helpers
+# ═══════════════════════════════════════════════════════════════════════
+
+
+def _build_orca_pka_settings(proton_index, shared, opt_settings):
```
xinglong-zhang: can be a @classmethod of ORCApKaJobSettings class since ORCApKaJobSettings is returned.

xinglong-zhang: then use as `pka_settings = ORCApKaJobSettings.build_orca_pka_settings(...)`

chemsmart/io/gaussian/output.py
```diff
@@ -2972,3 +2972,569 @@ def final_translation_vector(self):
...
+            concentration=concentration,
+            pressure=pressure,
+            cutoff_entropy_grimme=cutoff_entropy_grimme,
+            cutoff_enthalpy=cutoff_enthalpy,
+            entropy_method=entropy_method,
+            scheme=scheme,
+            delta_G_proton=delta_G_proton,
+        )
+
+    @classmethod
```
xinglong-zhang: should not be decorated with @classmethod, since it does not return the class object

xinglong-zhang: please check elsewhere and everywhere too

chemsmart/io/orca/output.py
```diff
@@ -3679,3 +3679,373 @@ def _get_ts_info(self):
...
+            concentration=concentration,
+            pressure=pressure,
+            cutoff_entropy_grimme=cutoff_entropy_grimme,
+            cutoff_enthalpy=cutoff_enthalpy,
+            entropy_method=entropy_method,
+            scheme=scheme,
+            delta_G_proton=delta_G_proton,
+        )
+
+    @classmethod
```
xinglong-zhang: should not have @classmethod

chemsmart/utils/utils.py
xinglong-zhang: IO is for input/output and not only for molecule object.

xinglong-zhang: utils are for small utility functions. maybe more appropriate to be in io.py

xinglong-zhang: maybe should be in chemsmart.io.file.py

chemsmart/utils/utils.py
```diff
@@ -2021,3 +2021,1467 @@ def token_score(token):
...
+    directory = directory or "."
+    for suffix in suffixes:
+        for ext in extensions:
+            candidate = os.path.join(directory, f"{basename}{suffix}{ext}")
+            if os.path.isfile(candidate):
+                return candidate
+    return os.path.join(directory, f"{basename}{suffixes[0]}{extensions[0]}")
+
+
+def discover_pka_target_companion_outputs(ha_gas_path, program=None):
```
xinglong-zhang: these utility functions may be in chemsmart.utils.io.py

xinglong-zhang (changes requested):
comments and requested changes for your consideration, @liu-jing-yi

chemsmart/cli/pka.py
liu-jing-yi: Removed.

chemsmart/utils/utils.py
liu-jing-yi: Updated.

chemsmart/cli/gaussian/pka.py
liu-jing-yi: Updated.

chemsmart/cli/orca/pka.py
liu-jing-yi: Refactored.

chemsmart/io/gaussian/output.py
liu-jing-yi: `print_pka_summary` `resolve_column` `resolve_column` `parse_table` and `compute_pka` were refactored as `@staticmethod`.

chemsmart/io/orca/output.py
liu-jing-yi: Fixed.

chemsmart/cli/gaussian/pka.py
liu-jing-yi: Updated.

liu-jing-yi:
> > ### CHEMSMART PKa Implementation (updated on Jun. 12, 2026)
> > Refactors the pKa workflow: job submission stays on Gaussian/ORCA (pka submit / batch); analysis is backend-independent via chemsmart run pka analyze / batch-analyze (mixed .log/.out supported).
> > 
> > 1. Highlights
> > 
> > 
> > * Proton exchange (default) and direct (-s direct) thermodynamic cycles with dual-level gas-phase freq + solvent SP
> > * CSV batch tables and multi-fragment ChemDraw CDXML (pka batch → one job per fragment, coloured-proton auto-detection)
> > * CDXML rows in CSV may omit proton_index; charge/multiplicity inferred from structures when not set on the CLI
> > * Cluster scripts reconstruct per row/fragment with explicit --proton-index, --index, --label, charge, and multiplicity
> > 
> > 
> > 2. Fixes
> > 
> > 
> > * Duplicate CDXML batch submissions
> > * Shared output filenames across fragments
> > * CSV table input incorrectly passed to molecule loading
> > 
> > 
> > 3. Example output (these PKa values are not physically meaningful, but are derived from a test reference acid)
> > 
> > ```
> > ==============================================================================
> > Batch pKa Results (Dual-level Proton Exchange)
> > ==============================================================================
> > Temperature: 278.15 K
> > Pressure: 1.0 atm
> > basename                              pKa   ΔG_soln (kcal/mol)
> > ------------------------------------------------------------------------------
> > pka_scale_frag1                     -8.17             -23.8913
> > pka_scale_frag4                     -1.24             -15.0654
> > pka_scale_frag3                      9.22              -1.7521
> > pka_scale_frag5                     -2.68             -16.8958
> > ==============================================================================
> > ```
> 
> @copilot, has these features been well-documented in the RTDs? If not, please update RTDs accordingly.

RTDS are updated.

chemsmart/cli/gaussian/pka.py
```diff
@@ -0,0 +1,525 @@
...
+        pka_kwargs.get("solvent_id"),
+        getattr(opt_settings, "solvent_id", None),
+        getattr(sp_settings, "solvent_id", None) if sp_settings else None,
+        "water",
+    )
+
+    pka_kwargs["solvent_model"] = solvent_model
+    pka_kwargs["solvent_id"] = solvent_id
+
+    return GaussianpKaJobSettings(
```
xinglong-zhang: Did I mentioned in earlier review that, since we are returning a GaussianpKaJobSettings object, this can be a @classmethod of `class GaussianpKaJobSettings` ? Any particular reason why this is in util?

chemsmart/io/file.py
```diff
@@ -481,52 +504,882 @@ def _preprocess_cdxml_remove_multi_attachments(filepath):
...
+      by higher-level pKa logic (for example, to build :class:`PKaMolecule`
+      instances or similar pKa-aware molecule objects).
+
+    The class does **not** perform the pKa calculation itself; instead it
+    focuses on converting CDXML drawings into a representation where
+    ionizable sites and their associated protons can be identified reliably.
+    """
+
+    @staticmethod
+    def resolve_proton_index(filename, proton_index, color_code=None):
```
xinglong-zhang: isn't the `filename` argument already in `self.filename` when you instantiate the PKaCDXFile object?

chemsmart/cli/gaussian/pka.py
```diff
@@ -0,0 +1,525 @@
...
+            "fragment_index": idx,
+            "label": label,
+        }
+        jobs.append(job)
+
+    logger.info(f"Created {len(jobs)} pKa jobs from multi-fragment CDXML")
+    return jobs
+
+
+def _log_pka_settings(pka_settings, proton_index, shared):
```
xinglong-zhang: This does not look like a private method, it should be simply `log_pka_settings()` and can be used elsewhere (presumably).

chemsmart/cli/orca/pka.py
```diff
@@ -0,0 +1,463 @@
...
+            "scheme": row_shared["scheme"],
+            "label": base_label,
+        }
+        jobs.append(job)
+
+    logger.info(f"Created {len(jobs)} ORCA pKa jobs from table")
+    return jobs
+
+
+def _create_orca_pka_jobs_from_molecules(
```
xinglong-zhang, line 403: should this be a private method - i.e., only called in this file by one class object and nothing else? If not, consider using public method, so that it can be imported and used elsewhere.

chemsmart/cli/orca/pka.py
```diff
@@ -0,0 +1,463 @@
...
+    if is_pka_cdxml_input(input_table_path):
+        return batch_pka_jobs_from_cdxml(
+            ctx,
+            skip_completed,
+            _create_orca_pka_jobs_from_molecules,
+            lambda ctx, **invoke_kwargs: ctx.invoke(submit, **invoke_kwargs),
+            **kwargs,
+        )
+
+    import copy
```
xinglong-zhang: import statements at the start of function (or start of file).

chemsmart/cli/pka.py
```diff
@@ -0,0 +1,1399 @@
...
+        "pressure": pressure,
+        "s_freq_cutoff": cutoff_entropy_grimme,
+        "h_freq_cutoff": cutoff_enthalpy,
+        "entropy_method": entropy_method,
+        "energy_units": "hartree",
+        "check_imaginary_frequencies": True,
+    }
+
+
+def _require_thermochemistry_property(thermo, filepath, attr, label):
```
xinglong-zhang: this method name seems to suggest a True/False return result, any better name?

chemsmart/cli/pka.py
```diff
@@ -0,0 +1,1399 @@
...
+    thermo_kwargs = _pka_thermochemistry_kwargs(
+        temperature,
+        concentration,
+        pressure,
+        cutoff_entropy_grimme,
+        cutoff_enthalpy,
+        entropy_method,
+    )
+
+    def get_species_thermo(file_path, name):
```
xinglong-zhang: use filepath consistently instead of file_path. check everywhere!

chemsmart/io/gaussian/output.py
```diff
@@ -2972,3 +2972,568 @@ def final_translation_vector(self):
...
+        """
+        Get the electronic energy (E) in specified units.
+
+        This is the raw SCF energy from the Gaussian calculation,
+        converted to the specified energy units.
+
+        Returns:
+            float: Electronic energy in specified units (default: hartree).
+        """
+        from chemsmart.utils.constants import energy_conversion
```
xinglong-zhang: this import statement is called inside each property, we can simply import it once at the beginning of the file.

chemsmart/io/gaussian/output.py
```diff
@@ -2972,3 +2972,568 @@ def final_translation_vector(self):
...
+                )
+                if thermo.qrrho_entropy_times_temperature
+                else None
+            ),
+            "gibbs_free_energy": self.gibbs_free_energy_in_units,
+            "qh_gibbs_free_energy": self.qh_gibbs_free_energy,
+        }
+
+    @classmethod
+    def from_settings(cls, filename, settings):
```
xinglong-zhang: interesting - is it possible to create a Gaussian16pKaOutput object from purely GaussianpKaJobSettings? will there be loss of information?

chemsmart/io/gaussian/output.py
```diff
@@ -2972,3 +2972,568 @@ def final_translation_vector(self):
...
+            temperature=temperature,
+            concentration=concentration,
+            pressure=pressure,
+            cutoff_entropy_grimme=cutoff_entropy_grimme,
+            cutoff_enthalpy=cutoff_enthalpy,
+            energy_units=energy_units,
+        )
+
+    @classmethod
+    def from_pka_settings(
```
xinglong-zhang: similarly

chemsmart/io/gaussian/output.py
```diff
@@ -2972,3 +2972,568 @@ def final_translation_vector(self):
...
+            concentration=concentration,
+            pressure=pressure,
+            cutoff_entropy_grimme=cutoff_entropy_grimme,
+            cutoff_enthalpy=cutoff_enthalpy,
+            entropy_method=entropy_method,
+            scheme=scheme,
+            delta_G_proton=delta_G_proton,
+        )
+
+    # Alias for backward compatibility
```
xinglong-zhang: Are these used at all in any version? seems like they are remnants from debugging process during code development? If so, they may be removed to avoid confusion.

chemsmart/io/orca/output.py
```diff
@@ -3679,3 +3679,371 @@ def _get_ts_info(self):
...
+                h_freq_cutoff=self.cutoff_enthalpy,
+                energy_units=self.energy_units,
+                check_imaginary_frequencies=True,
+            )
+        return self._thermochemistry
+
+    @property
+    def electronic_energy_in_units(self):
+        """Electronic energy (E) in specified units."""
+        from chemsmart.utils.constants import energy_conversion
```
xinglong-zhang: similarly, import once instead of multiple times.

chemsmart/io/orca/output.py
```diff
@@ -3679,3 +3679,371 @@ def _get_ts_info(self):
...
+            "gibbs_free_energy": self.gibbs_free_energy_in_units,
+            "qh_gibbs_free_energy": self.qh_gibbs_free_energy,
+        }
+
+    # ------------------------------------------------------------------
+    # Factory methods
+    # ------------------------------------------------------------------
+
+    @classmethod
+    def from_settings(cls, filename, settings):
```
xinglong-zhang: as previously. I assume that when you address Gaussian codes, you will address ORCA codes too, so i am less meticulous when checking ORCA codes

chemsmart/io/orca/output.py
```diff
@@ -3679,3 +3679,371 @@ def _get_ts_info(self):
...
+            concentration=concentration,
+            pressure=pressure,
+            cutoff_entropy_grimme=cutoff_entropy_grimme,
+            cutoff_enthalpy=cutoff_enthalpy,
+            entropy_method=entropy_method,
+            scheme=scheme,
+            delta_G_proton=delta_G_proton,
+        )
+
+    # Aliases for backward compatibility
```
xinglong-zhang: as for gaussian

chemsmart/io/file.py
```diff
@@ -481,52 +504,882 @@ def _preprocess_cdxml_remove_multi_attachments(filepath):
...
+        root = self._parse_cdxml_root()
+
+        list_of_elements = []
+        for fragment in self._iter_top_level_fragments(root):
+            list_of_elements.extend(self._parse_fragment_nodes(fragment, pt))
+
+        return list_of_elements
+
+    @staticmethod
+    def _find_parent(root, target):
```
xinglong-zhang: can be a general util method?

chemsmart/io/file.py
```diff
@@ -481,52 +504,882 @@ def _preprocess_cdxml_remove_multi_attachments(filepath):
...
+        """Return the parent element of *target* inside *root*, or None."""
+        for parent in root.iter():
+            for child in parent:
+                if child is target:
+                    return parent
+        return None
+
+    def _parse_cdxml_root(self):
```
xinglong-zhang, line 670: better as a util function `parse_cdxml_root(filename)`?

xinglong-zhang: I guess it may also be okay here..

xinglong-zhang: never mind the first comment - the more i read the methods, the more i realise they belong to operations on the object self, which is a CDXML file.

xinglong-zhang, line 670: consider using as public method?

chemsmart/io/file.py
```diff
@@ -481,52 +504,882 @@ def _preprocess_cdxml_remove_multi_attachments(filepath):
...
+        """Parse CDXML and return the XML root element."""
+        try:
+            tree = ET.parse(self.filename)
+        except ET.ParseError as exc:
+            raise ValueError(
+                f"Failed to parse CDXML file {self.filename}: {exc}"
+            ) from exc
+        return tree.getroot()
+
+    def _iter_top_level_fragments(self, root):
```
xinglong-zhang, line 680: consider using as public method, if this will be called from the object, but if it is **only** called by other class method/function, then private method is okay.

chemsmart/io/file.py
```diff
@@ -481,52 +504,882 @@ def _preprocess_cdxml_remove_multi_attachments(filepath):
...
+        atoms = self.parse_cdxml_element_colors()
+
+        if not atoms:
+            raise ValueError(f"No atoms found in CDXML file: {self.filename}")
+
+        if color_code is not None:
+            return self._proton_index_by_color(atoms, color_code)
+        return self._proton_index_auto(atoms)
+
+    # ----- private helpers for get_colored_proton_index -----
```
xinglong-zhang, line 868: this is good and clear annotations.

chemsmart/utils/io.py
```diff
@@ -31,6 +33,151 @@
 
 logger = logging.getLogger(__name__)
 
+
+class TabularDataset:
```
xinglong-zhang: should this be under io? perhaps we can have another file called io.tables.py in the same level as io.file.py?

I imagine there may be similar such objects in future, esp for high-throughput workflows. Or maybe we can put it in io.datasets.py?

chemsmart/utils/utils.py
```diff
@@ -2021,3 +2021,1310 @@ def token_score(token):
...
 
     return " ".join(new_tokens)
+
+
+# ---------------------------------------------------------------------------
+# pKa Table Parsing Utilities
+# ---------------------------------------------------------------------------
+
+
+def normalize_table_cell(value):
```
xinglong-zhang: Maybe should be under utils.io.py? or can create a file called utils.datasets.py corresponding to io.datasets.py or utils.tables.py corresponding to io.tables.py? useful if we anticipate more similar dataset/table objects in future.

chemsmart/utils/utils.py
```diff
@@ -2021,3 +2021,1310 @@ def token_score(token):
...
+            return None
+    except (TypeError, ImportError):
+        pass
+    if isinstance(value, str):
+        stripped = value.strip()
+        return stripped if stripped else None
+    return value
+
+
+class PKaTableEntry:
```
xinglong-zhang: similarly, maybe can move all table related to io.tables.py

xinglong-zhang (changes requested):
Things are looking great! We should be able to merge after some minor changes and rigourous usage testing. @liu-jing-yi

chemsmart/cli/orca/pka.py
liu-jing-yi: Should be okay to keep it private, as this method is only called in `chemsmart/cli/orca/pka.py`.

chemsmart/cli/orca/pka.py
liu-jing-yi: Moved to start of the file.

chemsmart/cli/gaussian/pka.py
liu-jing-yi: `build_gaussian_pka_settings` was refactored to be a @classmethod of `GaussianpKaJobSettings`.

chemsmart/cli/gaussian/pka.py
liu-jing-yi: Renamed.

chemsmart/io/file.py
liu-jing-yi: Not really. `resolve_proton_index` is rather a helper that decides whether to open a CDXML file (for subsequent auto-detection of proton), which can run without really instantiating a `PKaCDXFile` object.

chemsmart/cli/pka.py
liu-jing-yi: Refactored to `_extract_thermochemistry_property`.

chemsmart/cli/pka.py
liu-jing-yi: Renamed.

chemsmart/io/gaussian/output.py
liu-jing-yi: Updated.

chemsmart/io/gaussian/output.py
liu-jing-yi: Not exactly. `Gaussian16pKaOutput` cannot be built from `GaussianpKaJobSettings` alone. Because `GaussianpKaJobSettings` only supply thermochemistry parameters, but the output class still needs a completed log file for further processing. The docstring has been updated accordingly.

chemsmart/io/gaussian/output.py
liu-jing-yi: Similarly, we cannot build output objects from `GaussianpKaJobSettings` alone. It must pass completed Gaussian log paths.

chemsmart/io/gaussian/output.py
liu-jing-yi: Yes. This has been removed.

chemsmart/io/orca/output.py
liu-jing-yi: Updated.

chemsmart/io/orca/output.py
liu-jing-yi: Removed.

chemsmart/io/file.py
liu-jing-yi: At this stage, it'd better to keep it private, as it is only used in PKaCDXFile. For future use, we may consider syntax like: 
```
@property
def cdxml_root(self):  
    if self._root is None:
        self._root = self._parse_cdxml_root()
    return self._root
```

chemsmart/utils/io.py
liu-jing-yi: That's a good point. `TabularDataset` has been moved to a new `chemsmart/io/datasets.py` file to leave room for future high-throughput dataset types.

chemsmart/io/file.py
liu-jing-yi: Refactored to be general module helper.

chemsmart/utils/utils.py
liu-jing-yi: Updated.

chemsmart/utils/utils.py
liu-jing-yi: updated.

chemsmart/io/orca/output.py
liu-jing-yi: ORCA codes were updated to keep consistency with Gaussian.

chemsmart/io/file.py
liu-jing-yi: Better to keep it private as it is only called inside PKaCDXFile.

chemsmart/io/gaussian/output.py
xinglong-zhang: IN this case, it should not be a @classmethod, since this will return the Gaussian16pKaOutput object, if i then want to access the normal attributes in this object, it cannot be found, if it is built from settings instead of from the log file.

chemsmart/io/gaussian/output.py
xinglong-zhang: Thus should not be @classmethod

xinglong-zhang: I don't understand why you need to retrospectively construct output file object from the settings?

chemsmart/io/file.py
xinglong-zhang: when you call `PKaCDXFile.resolve_proton_index(filename)`, isn't this filename the same as `PKaCDXFile(filename).resolve_proton_index()`?

chemsmart/utils/io.py
```diff
@@ -367,6 +373,76 @@ def get_program_type_from_file(filepath):
     return "unknown"
 
 
+def detect_program_type_from_files(filepaths, allowed_programs=None):
```
xinglong-zhang: Is there a need to write this? I remember Huiwen has a utility function for this already when dealing with thermochemistry

xinglong-zhang (changes requested):
@liu-jing-yi, just a few more comments to consider and we can merge today

xinglong-zhang:
@liu-jing-yi, as mentioned before, just note that a class method (a function inside a class object) is different from @classmethod that will create the class object from this definition.

chemsmart/utils/io.py
liu-jing-yi: This seems to be some redundant lines during previous refactoring. Removed.

liu-jing-yi:
@xinglong-zhang For the `from_settings` / `from_pka_setting`s comments: both were redundant wrappers that still required log paths and only unpacked thermochemistry settings. Removed them; the job layer now uses `for_pka_species` directly.

For the comment on `resolve_proton_index`: this static wrapper also handles cases where no file object is needed (explicit -pi) or where the input isn’t CDX (validation/errors). However, placing this method in `PKaCDXFile` is misleading, because it can accept other file formats than cdxml files.  I've moved `resolve_proton_index `to `chemsmart/cli/pka.py`, alongside `is_pka_cdxml_input`. `PKaCDXFile` now only handles CDX/CDXML parsing.

chemsmart/io/gaussian/output.py
```diff
@@ -2972,3 +2973,490 @@ def final_translation_vector(self):
...
+            ref_file=ref_file,
+            temperature=temperature,
+            concentration=concentration,
+            pressure=pressure,
+            cutoff_entropy_grimme=cutoff_entropy_grimme,
+            cutoff_enthalpy=cutoff_enthalpy,
+            energy_units=energy_units,
+        )
+
+    @classmethod
```
xinglong-zhang: since this returns a list of Gaussian16pKaOutput object and is not the cls itself, perhaps this should just be a utility function, e.g., `create_list_of_pka_outputs_from_files()` instead?

xinglong-zhang:
@liu-jing-yi, one last issue to address and i will merge, thanks.

chemsmart/io/gaussian/output.py
xinglong-zhang: According to GPT, this is okay and I misunderstood:

> it is okay for a @classmethod to create and return multiple class objects instead of a single object. Python does not require a classmethod to return cls(...); it can return any useful object: one instance, a list, a dict, a tuple, etc.
> 
> In your case, for_pka_species is reasonable because it uses cls(...), so subclasses of ORCApKaOutput would also work correctly.

xinglong-zhang (approved):

chemsmart/analysis/thermochemistry.py
```diff
@@ -137,6 +137,11 @@ def __init__(
             else None
         )
 
+    @classmethod
+    def from_filepath(cls, filepath, **kwargs):
+        """Construct a Thermochemistry object from an output filepath."""
+        return cls(filename=filepath, **kwargs)
+
```
Huiwen-Tan, line 144: Why is this needed? @liu-jing-yi

liu-jing-yi: Addressed in PR #604

PR #569, opened 2026-05-22
feat: agent for chemsmart

xinglong-zhang:


xinglong-zhang:
@Hongjiseung-ROK, please open a PR to my `agent` branch instead of `main`, so that people can start using and testing it before it is merged into `main`. Thanks!

PR #586, opened 2026-06-13
Database Stability Improvements

Huiwen-Tan:


chemsmart/database/assemble.py
```diff
@@ -99,7 +99,7 @@ def assemble(self):
         route_tokens = canonicalize_route_string(
             self.output.route_string,
             drop_terms=[
-                self.output.functional,
+                self.output.method,
```
xinglong-zhang, line 102: why is functional changed to method? Does method also include ab inito methods? What about semi-empircal,e.g., PM6, methods?

chemsmart/database/inspect.py
```diff
@@ -124,10 +124,11 @@ def overview(self):
             stats["solvated_records"] = row["solvated_records"]
             stats["gas_phase_records"] = row["gas_phase_records"]
 
-            # Frequency and solvent count (from record_structures table)
-            stats["freq_count"] = conn.execute(
-                "SELECT COUNT(DISTINCT record_id) FROM record_structures WHERE vibrational_frequencies_json IS NOT NULL"
+            # Forces count (records that have at least one structure with forces)
```
xinglong-zhang, line 127: why is the original freq_count removed?

chemsmart/database/assemble.py
xinglong-zhang: I saw this in later file. Again, do we also need to include semi-empirical one?

xinglong-zhang (changes requested):
Thanks for the revision, please consider the comments @Huiwen-Tan

chemsmart/database/assemble.py
Huiwen-Tan: The change from `functional` to `method` was intentional so that the calculation method can also be correctly identified for non-DFT calculations. For Gaussian, `method` already covers DFT functionals, ab initio methods, and semi-empirical methods (`self.functional or self.ab_initio or self.semiempirical`). For ORCA, `method` currently only covers DFT functionals and ab initio methods (`self.functional or self.ab_initio`). Semi-empirical methods are not included yet, but I agree they should be. I'll try to add semi-empirical method detection to the ORCA parser as well.

chemsmart/database/inspect.py
Huiwen-Tan: I removed the original `freq_count` mainly because it was derived from deduplication statistics on `record_structures.vibrational_frequencies_json` (structure-level → distinct records)—an indirect metric that occasionally led to underestimation. The `thermo_count` I now use comes from `records.gibbs_free_energy IS NOT NULL`, which is a direct record‑level field statistic—stable and consistent with our thermochemistry logic.

xinglong-zhang (approved):

xinglong-zhang:
@Huiwen-Tan, please follow up with copilot comments at your convenience.

chemsmart/database/inspect.py
```diff
@@ -124,10 +124,11 @@ def overview(self):
...
+            stats["forces_count"] = conn.execute(
+                "SELECT COUNT(DISTINCT record_id) FROM record_structures WHERE forces_json IS NOT NULL"
             ).fetchone()[0]
+            # Solvent count
             rows = conn.execute("""
```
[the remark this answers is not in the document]
Huiwen-Tan: Thanks for the suggestion. However, `records.gibbs_free_energy` is set if and only if `self.output.freq` is `True` in the assembler. This directly reflects whether the job was a frequency calculation.
`COUNT(DISTINCT record_id) FROM record_structures WHERE vibrational_frequencies_json IS NOT NULL` is an indirect, structure-level count that can undercount records compared to the record-level `gibbs_free_energy` signal. The current approach using `gibbs_free_energy IS NOT NULL` is the more stable and semantically correct way for "this was a frequency job".

chemsmart/database/inspect.py
```diff
@@ -238,12 +239,10 @@ def format_overview(self):
...
         lines.append(
-            format_kv("Records with Frequencies", f"{stats['freq_count']}")
+            format_kv("Records with Forces", f"{stats['forces_count']}")
         )
         lines.append(
-            format_kv(
-                "Records with Thermochemistry", f"{stats['thermo_count']}"
-            )
+            format_kv("Records with Frequencies", f"{stats['thermo_count']}")
         )
```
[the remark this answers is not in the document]
Huiwen-Tan: Same reason as above.

PR #588, opened 2026-06-26
xTB I/O, thermochemistry, database

Huiwen-Tan:


xinglong-zhang:
@Huiwen-Tan, looks like one of the tests is failing on windows platform, can you (or ask an intern to) fix this?

Huiwen-Tan:
> @Huiwen-Tan, looks like one of the tests is failing on windows platform, can you (or ask an intern to) fix this?

okay, I fixed it in commit https://github.com/xinglong-zhang/chemsmart/pull/588/changes/d910ac8ee4f8f5aaa3a124e5db8d99b1bef5a712.

chemsmart/utils/repattern.py
```diff
@@ -127,6 +127,10 @@
     r"\* Starting time:\s+(\w{3} \w{3}\s+\d+ \d{2}:\d{2}:\d{2} \d{4})"
 )
 
+xtb_date_pattern = (
+    r"\* finished run on (\d{4}/\d{2}/\d{2}) at (\d{2}:\d{2}:\d{2})"
+)
```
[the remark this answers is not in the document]
Huiwen-Tan: Thanks for flagging this, but the current pattern already works with fractional seconds in our test data (`tests/data/XTBTests/outputs/...`). Python's re.search() partially matches `\d{2}:\d{2}:\d{2}` against `20:39:51.283`, capturing 20:39:51 in group 2 without requiring end-of-string. That value parses correctly with `strptime(..., "%Y/%m/%d %H:%M:%S")`.

The suggested change would capture `20:39:51.283` in group 2, which then fails strptime with `"%Y/%m/%d %H:%M:%S"` (unconverted data remains: `.283`), causing `file_date` to return `None` for all XTB outputs. I verified this on all 7 test `.out` files: original pattern succeeds, suggested pattern fails on all of them.

If we ever want sub-second precision, we'd need to update both the regex and the `strptime` format (e.g. `"%Y/%m/%d %H:%M:%S.%f"`), but truncating to seconds is intentional for provenance `source_date`.

chemsmart/utils/mixins.py
```diff
@@ -1409,6 +1417,197 @@ def read_settings(self):
...
+                match = re.search(xtb_date_pattern, line)
+                if match:
+                    date_str = match.group(1)  # YYYY/MM/DD
+                    time_str = match.group(2)  # HH:MM:SS
+                    try:
+                        # Convert from YYYY/MM/DD HH:MM:SS to YYYY-MM-DD HH:MM:SS
+                        date_obj = datetime.strptime(
+                            f"{date_str} {time_str}", "%Y/%m/%d %H:%M:%S"
+                        )
+                        return date_obj.strftime("%Y-%m-%d %H:%M:%S")
```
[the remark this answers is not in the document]
Huiwen-Tan: Applying both changes together would work but is unnecessary: it would capture fractional seconds only to discard them again via strftime("%Y-%m-%d %H:%M:%S"). Second-precision provenance dates are intentional — no change needed.

chemsmart/io/molecules/structure.py
```diff
@@ -1234,6 +1267,24 @@ def _read_orca_outfile(filepath, index):
...
+            filepath (str): Path to xTB calculation directory
+            index (str or int): Index for multi-structure files
+
+        Returns:
+            Molecule: Molecule object from xTB output
+        """
+        from chemsmart.io.xtb.output import XTBOutput
+
+        xtb_output = XTBOutput(folder=filepath)
+        return xtb_output.get_molecule(index=index)
```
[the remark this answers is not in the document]
Huiwen-Tan: Great catch — thanks! With this fix, callers can pass the path to an xTB main output file (e.g. `water_ohess.out`) and we still discover the full calculation directory to build the `Molecule` via `XTBOutput`. That compatibility matters for many file-input-only components — e.g. `chemsmart run mol`, `chemsmart sub gaussian`, and `chemsmart sub orca` — and makes it much easier to reuse xTB results as downstream job input.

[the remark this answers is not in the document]
Huiwen-Tan: fixed in commit https://github.com/xinglong-zhang/chemsmart/pull/588/changes/32a36fccd31756f301c0a0de0afa1434f85ed018.

chemsmart/analysis/thermochemistry.py
```diff
@@ -62,7 +64,8 @@ class Thermochemistry:
 
     def __init__(
         self,
-        filename,
+        filename=None,
```
xinglong-zhang: For Thermochemistry class, it is better to do this: https://chatgpt.com/share/6a5b6971-53e4-83ec-a88c-104d0b39b777

I think what I told Jingyi to do last time is better - to be able to create a thermochemistry object from_filepath and from_folder classmethods

xinglong-zhang:
@Huiwen-Tan, can you fix this design of Thermochemistry class

xinglong-zhang:
`Thermochemistry` should primarily represent the data and settings required for a thermochemical calculation, rather than how that data was located on disk.

Using `filename` and `folder` in `__init__ `creates several problems:

- The constructor must decide which source takes precedence.
- It couples thermochemistry calculations to filesystem parsing.
- It becomes difficult to construct the object from an existing Molecule, parsed logfile, database record, or in-memory data.
- filename and folder are not meaningful properties of thermochemistry itself.

A cleaner design would be:
```
class Thermochemistry:
    """Calculate thermochemical quantities from molecular data."""

    def __init__(
        self,
        molecule,
        temperature=298.15,
        concentration=None,
        pressure=1.0,
        *,
        use_weighted_mass=True,
        alpha=4,
        s_freq_cutoff=100.0,
        h_freq_cutoff=100.0,
        energy_units="hartree",
        check_imaginary_frequencies=True,
        rotational_mode="physical",
    ):
        self.molecule = molecule
        self.temperature = temperature
        self.concentration = concentration
        self.pressure = pressure
```

Then provide alternate constructors:
```

from pathlib import Path


class Thermochemistry:
    # __init__ omitted

    @classmethod
    def from_filepath(cls, filepath, **kwargs):
        molecule = Molecule.from_filepath(filepath)

        return cls(
            molecule=molecule,
            **kwargs,
        )

    @classmethod
    def from_folder(cls, folder, *, pattern="*.log", **kwargs):
        folder = Path(folder)
        filepaths = sorted(folder.glob(pattern))

        if not filepaths:
            raise FileNotFoundError(
                f"No files matching {pattern!r} found in {folder}"
            )

        return [
            cls.from_filepath(filepath, **kwargs)
            for filepath in filepaths
        ]
```

I would also reconsider `from_folder()`: if a folder contains multiple calculations, returning a single Thermochemistry object is ambiguous. It should either:

- return list[Thermochemistry];
- return a dedicated ThermochemistryCollection; or
- require a clear file-selection rule.

Also, outputfile probably does not belong in the constructor. Saving is a separate responsibility and could instead be:

`thermo.write("thermochemistry.csv")`

So the recommended usage becomes:

```
thermo = Thermochemistry.from_filepath(
    "calculation.log",
    temperature=393.15,
    rotational_mode="physical",
)
```

or, when the molecule is already parsed:

```
thermo = Thermochemistry(
    molecule,
    temperature=393.15,
    rotational_mode="physical",
)
```

This separation will make Thermochemistry much easier to test and reuse.

xinglong-zhang:
> `Thermochemistry` should primarily represent the data and settings required for a thermochemical calculation, rather than how that data was located on disk.
> 
> Using `filename` and `folder` in `__init__ `creates several problems:
> 
> * The constructor must decide which source takes precedence.
> * It couples thermochemistry calculations to filesystem parsing.
> * It becomes difficult to construct the object from an existing Molecule, parsed logfile, database record, or in-memory data.
> * filename and folder are not meaningful properties of thermochemistry itself.
> 
> A cleaner design would be:
> 
> ```
> class Thermochemistry:
>     """Calculate thermochemical quantities from molecular data."""
> 
>     def __init__(
>         self,
>         molecule,
>         temperature=298.15,
>         concentration=None,
>         pressure=1.0,
>         *,
>         use_weighted_mass=True,
>         alpha=4,
>         s_freq_cutoff=100.0,
>         h_freq_cutoff=100.0,
>         energy_units="hartree",
>         check_imaginary_frequencies=True,
>         rotational_mode="physical",
>     ):
>         self.molecule = molecule
>         self.temperature = temperature
>         self.concentration = concentration
>         self.pressure = pressure
> ```
> 
> Then provide alternate constructors:
> 
> ```
> 
> from pathlib import Path
> 
> 
> class Thermochemistry:
>     # __init__ omitted
> 
>     @classmethod
>     def from_filepath(cls, filepath, **kwargs):
>         molecule = Molecule.from_filepath(filepath)
> 
>         return cls(
>             molecule=molecule,
>             **kwargs,
>         )
> 
>     @classmethod
>     def from_folder(cls, folder, *, pattern="*.log", **kwargs):
>         folder = Path(folder)
>         filepaths = sorted(folder.glob(pattern))
> 
>         if not filepaths:
>             raise FileNotFoundError(
>                 f"No files matching {pattern!r} found in {folder}"
>             )
> 
>         return [
>             cls.from_filepath(filepath, **kwargs)
>             for filepath in filepaths
>         ]
> ```
> 
> I would also reconsider `from_folder()`: if a folder contains multiple calculations, returning a single Thermochemistry object is ambiguous. It should either:
> 
> * return list[Thermochemistry];
> * return a dedicated ThermochemistryCollection; or
> * require a clear file-selection rule.
> 
> Also, outputfile probably does not belong in the constructor. Saving is a separate responsibility and could instead be:
> 
> `thermo.write("thermochemistry.csv")`
> 
> So the recommended usage becomes:
> 
> ```
> thermo = Thermochemistry.from_filepath(
>     "calculation.log",
>     temperature=393.15,
>     rotational_mode="physical",
> )
> ```
> 
> or, when the molecule is already parsed:
> 
> ```
> thermo = Thermochemistry(
>     molecule,
>     temperature=393.15,
>     rotational_mode="physical",
> )
> ```
> 
> This separation will make Thermochemistry much easier to test and reuse.

@Huiwen-Tan

xinglong-zhang (changes requested):

Huiwen-Tan:
> `Thermochemistry` should primarily represent the data and settings required for a thermochemical calculation, rather than how that data was located on disk.
> 
> Using `filename` and `folder` in `__init__ `creates several problems:
> 
> * The constructor must decide which source takes precedence.
> * It couples thermochemistry calculations to filesystem parsing.
> * It becomes difficult to construct the object from an existing Molecule, parsed logfile, database record, or in-memory data.
> * filename and folder are not meaningful properties of thermochemistry itself.
> 
> A cleaner design would be:
> 
> ```
> class Thermochemistry:
>     """Calculate thermochemical quantities from molecular data."""
> 
>     def __init__(
>         self,
>         molecule,
>         temperature=298.15,
>         concentration=None,
>         pressure=1.0,
>         *,
>         use_weighted_mass=True,
>         alpha=4,
>         s_freq_cutoff=100.0,
>         h_freq_cutoff=100.0,
>         energy_units="hartree",
>         check_imaginary_frequencies=True,
>         rotational_mode="physical",
>     ):
>         self.molecule = molecule
>         self.temperature = temperature
>         self.concentration = concentration
>         self.pressure = pressure
> ```
> 
> Then provide alternate constructors:
> 
> ```
> 
> from pathlib import Path
> 
> 
> class Thermochemistry:
>     # __init__ omitted
> 
>     @classmethod
>     def from_filepath(cls, filepath, **kwargs):
>         molecule = Molecule.from_filepath(filepath)
> 
>         return cls(
>             molecule=molecule,
>             **kwargs,
>         )
> 
>     @classmethod
>     def from_folder(cls, folder, *, pattern="*.log", **kwargs):
>         folder = Path(folder)
>         filepaths = sorted(folder.glob(pattern))
> 
>         if not filepaths:
>             raise FileNotFoundError(
>                 f"No files matching {pattern!r} found in {folder}"
>             )
> 
>         return [
>             cls.from_filepath(filepath, **kwargs)
>             for filepath in filepaths
>         ]
> ```
> 
> I would also reconsider `from_folder()`: if a folder contains multiple calculations, returning a single Thermochemistry object is ambiguous. It should either:
> 
> * return list[Thermochemistry];
> * return a dedicated ThermochemistryCollection; or
> * require a clear file-selection rule.
> 
> Also, outputfile probably does not belong in the constructor. Saving is a separate responsibility and could instead be:
> 
> `thermo.write("thermochemistry.csv")`
> 
> So the recommended usage becomes:
> 
> ```
> thermo = Thermochemistry.from_filepath(
>     "calculation.log",
>     temperature=393.15,
>     rotational_mode="physical",
> )
> ```
> 
> or, when the molecule is already parsed:
> 
> ```
> thermo = Thermochemistry(
>     molecule,
>     temperature=393.15,
>     rotational_mode="physical",
> )
> ```
> 
> This separation will make Thermochemistry much easier to test and reuse.

Thanks — I agree with the separation-of-responsibilities concern. I addressed the filesystem ambiguity in the following ways:

Removed `folder` from `Thermochemistry.__init__`; `filename` is now the only input source.
Removed folder-based construction. Batch directory discovery is now handled by the CLI/job layer rather than `Thermochemistry`.
Added xTB single-file support: the main output file identifies the calculation directory, and `XTBOutput` loads the associated auxiliary files automatically.
Removed the output destination from the constructor state. The job layer now owns the output path, while serialization is handled separately by `log_results_to_file()`.

I have not yet changed the constructor to accept only `Molecule`. At present, a `Molecule` alone is insufficient for thermochemistry calculations because they also require source-level information such as vibrational frequencies, electronic energy, multiplicity, job type, rotational symmetry number, and, for `mode="gaussian"`, the printed rotational constants. These are currently provided by the parsed output object.

Therefore, changing the constructor to accept only `Molecule` would either lose required information or simply move the filesystem coupling somewhere less explicit.

For this change, I kept file-based parsing as the supported construction path, while removing the ambiguous `filename`/`folder` dual interface and moving directory discovery and output management out of the calculation class.

Huiwen-Tan:
A note on directory discovery and traversal semantics:

CHEMSMART treats Gaussian/ORCA and xTB differently at the filesystem level because their natural calculation units differ:

- A Gaussian/ORCA calculation is represented primarily by one output file.
- An xTB calculation is represented by a directory containing one main output plus auxiliary files such as `charges`, `xtbopt.log`, and `g98.out`.

### Thermochemistry

Thermochemistry directory processing is intentionally non-recursive:

| Program | `-d` behavior |
|---|---|
| Gaussian/ORCA | Process matching output files directly inside the specified directory |
| xTB | Process valid xTB calculation directories directly below the specified parent directory |

For example:

```text
calculations/
├── gaussian_a.log
├── gaussian_b.log
├── xtb_a/
│   ├── xtb_a.out
│   ├── charges
│   └── xtbopt.log
└── xtb_b/
    ├── xtb_b.out
    ├── charges
    └── xtbopt.log
```

Conceptually, `gaussian_a.log` and `xtb_a/` occupy the same batch level: one is a file-based calculation and the other is a folder-based calculation.

Nested xTB calculation directories are not searched by thermochemistry. This keeps `-d` predictable and bounded, matching the existing non-recursive Gaussian/ORCA behavior. For one xTB calculation, the user passes its main output explicitly:

```bash
chemsmart run thermochemistry -T 298.15 -f xtb_a/xtb_a.out
```

The main output is used to locate the calculation directory, and `XTBOutput` reads the associated auxiliary files.

### Database assembly

Database assembly has different semantics because its purpose is project-wide ingestion. It therefore scans recursively to arbitrary depth:

| Program | Database discovery |
|---|---|
| Gaussian/ORCA | Recursively discover output files |
| xTB | Recursively discover xTB main outputs and treat each output’s parent as a candidate calculation directory |

Each xTB candidate is validated with `is_xtb_calculation_directory`. A valid xTB directory must contain exactly one main output and the required directory markers. Ambiguous or invalid directories are logged and skipped, while assembly continues with other valid calculations.

This distinction is intentional:

- **Thermochemistry** performs a controlled batch operation at one user-selected level.
- **Database assembly** indexes an entire project tree.
- **Single-file analysis** remains consistent across Gaussian, ORCA, and xTB through `-f`.
- **xTB directory validation** prevents auxiliary files from different calculations from being combined accidentally.

chemsmart/io/xtb/folder.py
```diff
@@ -0,0 +1,221 @@
...
+        ├─ vibspectrum      → vibrational frequencies and IR intensities
+        ├─ wbo              → Wiberg bond orders
+        ├─ xtbopt.*         → optimized geometry (XYZ, SDF, Turbomole coord, PDB, VASP POSCAR,
+        │                     DFTB+ gen or Gaussian External format)
+        ├─ xtbtopo.mol      → molecular topology and bonding information
+        └─ ...              → other auxiliary files
+    """
+
+    PARSEABLE_GEOMETRY_EXTENSIONS = (".xyz", ".sdf")
+    UNSUPPORTED_GEOMETRY_EXTENSIONS = (
```
xinglong-zhang, line 35: .pdb should already be supported?

chemsmart/io/xtb/folder.py
```diff
@@ -0,0 +1,221 @@
...
+        # Check if unsupported format exists
+        for ext in self.UNSUPPORTED_GEOMETRY_EXTENSIONS:
+            geometry_files = self.get_all_files_in_current_folder_by_suffix(
+                ext
+            )
+            for filepath in geometry_files:
+                basename = os.path.basename(filepath)
+                if not basename.startswith("xtbopt"):
+                    logger.warning(
+                        f"Found input geometry file {filepath}, but format {ext} "
```
xinglong-zhang, line 169: use CHEMSMART.

chemsmart/io/xtb/folder.py
```diff
@@ -0,0 +1,221 @@
...
+
+        XTB outputs the optimized geometry in the same format as the input.
+        This method returns a parseable xtbopt.* file if available, otherwise
+        falls back to an unsupported format with a warning.
+
+        Search priority (parseable formats only):
+            1. xtbopt.xyz     → XYZ format (supported)
+            2. xtbopt.sdf     → SDF format (supported)
+
+        Unsupported formats (will show warning):
```
xinglong-zhang, line 189: we can add a note that these will be supported in future release.

chemsmart/cli/gaussian/gaussian.py
```diff
@@ -593,7 +593,26 @@ def gaussian(
     elif filename.endswith((".com", ".gjf", ".inp", ".out", ".log")):
         # filename supplied - we would want to use the settings from here
         #  and do not use any defaults!
-        job_settings = GaussianJobSettings.from_filepath(filename)
+        if filename.endswith(".out") and (
```
xinglong-zhang, line 596: nice!

chemsmart/cli/thermochemistry/thermochemistry.py
```diff
@@ -218,6 +220,14 @@ def thermochemistry(
     -o thermochemistry_results.dat`
     will compute thermochemistry for all Gaussian output files in the specified
     directory and save to `thermochemistry_results.dat`.
+
+    `chemsmart run thermochemistry -f water_opt/water_opt.out -T 298.15`
+    will compute thermochemistry for one xTB output file.
+
+    `chemsmart run thermochemistry -d xtb_calculations/ -p xtb -T 298.15
```
xinglong-zhang, line 227: in the docstrings and RTDs, can you talk a bit about the restrictions on the xtb_calculations/ folder? can it have different depths of xtb calcs?

chemsmart/database/utils.py
```diff
@@ -1,6 +1,7 @@
 import hashlib
```
xinglong-zhang, line 1: should be in `chemsmart.utils.database.py`? There is already `chemsmart.utils.dataset.py`, i think these two can be merged as one.

Also I think database should be under` chemsmart.jobs `just like everything else?

chemsmart/io/xtb/file.py
```diff
@@ -0,0 +1,1707 @@
...
+
+    Args:
+        filename (str): Path to the xTB output file (e.g., xtb.out, water_ohess.out)
+    """
+
+    def __init__(self, filename):
+        self.filename = filename
+
+    @property
+    def xtb_version(self):
```
xinglong-zhang: there is already `version` property in XTBFileMixin, why do you need this one more time?

xinglong-zhang: it's to distinguish from crest later?

chemsmart/io/xtb/file.py
```diff
@@ -0,0 +1,1707 @@
...
+        for i, line in enumerate(self.contents):
+            if line.startswith("molecular quadrupole (traceless):"):
+                for j_line in self.contents[i + 1 : i + 9]:
+                    quadrupole_lines.append(j_line)
+        if len(quadrupole_lines) == 0:
+            return None
+        return quadrupole_lines
+
+    @property
+    def qonly_molecular_quadrupole(self):
```
xinglong-zhang: may be better to call molecular_quadrupole_qonly and molecular_quadrupole_q_dip, so that if i only know i want to access molecular_quadrupole, i can type and search more easily

chemsmart/io/xtb/file.py
```diff
@@ -0,0 +1,1707 @@
...
+                if line.startswith("q+dip:"):
+                    xx, xy, yy, xz, yz, zz = [
+                        float(x) for x in line.split()[-6:]
+                    ]
+                    # convert to tensor
+                    return np.array([[xx, xy, xz], [xy, yy, yz], [xz, yz, zz]])
+        return None
+
+    @property
+    def full_molecular_quadrupole(self):
```
xinglong-zhang: as previously molecular_quadrupole_full might be better

chemsmart/io/xtb/file.py
```diff
@@ -0,0 +1,1707 @@
...
+        """Actual dipole moment including both charge distribution
+        and electronic polarization contributions"""
+        if self.molecular_dipole_lines is not None:
+            for line in self.molecular_dipole_lines:
+                if line.startswith("full:"):
+                    return np.array([float(x) for x in line.split()[1:4]])
+        return None
+
+    @property
+    def total_molecular_dipole_moment(self):
```
xinglong-zhang, line 592: this is fine, with total in front

chemsmart/io/xtb/file.py
```diff
@@ -0,0 +1,1707 @@
...
+            if line.startswith("molecular dipole:"):
+                for j_line in self.contents[i + 2 : i + 4]:
+                    # only get two lines
+                    dipole_lines.append(j_line)
+        if len(dipole_lines) == 0:
+            return None
+        return dipole_lines
+
+    @property
+    def qonly_molecular_dipole(self):
```
xinglong-zhang: see comment for molecular_quadrupole

chemsmart/io/xtb/file.py
```diff
@@ -0,0 +1,1707 @@
...
+        """Total enthalpy in Eh"""
+        return self._extract_final_information("TOTAL ENTHALPY")
+
+    @property
+    def gibbs_free_energy(self):
+        """Total free energy in Eh"""
+        return self._extract_final_information("TOTAL FREE ENERGY")
+
+    @property
+    def alpha_occ_eigenvalues(self):
```
xinglong-zhang, line 1050: I think xTB can do radicals too, just e.g., `--uhf=1` for one unpaired electron etc

chemsmart/io/xtb/file.py
```diff
@@ -0,0 +1,1707 @@
...
+    def thermal_gibbs_free_energy_correction(self):
+        return None
+
+    @property
+    def internal_energy(self):
+        return None
+
+    @property
+    def electronic_entropy(self):
+        # TODO
```
xinglong-zhang: TODO as in for the next PR?

chemsmart/io/xtb/file.py
```diff
@@ -0,0 +1,1707 @@
...
+                # Validate we have the correct number of gradient components
+                if len(grad_data) == 3 * self.num_atoms:
+                    gradient_array = np.array(grad_data).reshape(
+                        (self.num_atoms, 3)
+                    )
+                    return gradient_array
+        return None
+
+
+class XTBG98File(FileMixin):
```
xinglong-zhang, line 1515: Do you need to parse this again? Can't you set this file as a Gaussian16OutputFile object and read the relevant information? If so, it's better not to repeat the codes.

chemsmart/io/xtb/file.py
```diff
@@ -0,0 +1,1707 @@
...
+        atom1_index  atom2_index  bond_order_value
+
+    Args:
+        filename (str): Path to the wbo file
+    """
+
+    def __init__(self, filename):
+        self.filename = filename
+
+    # TODO: Add parsing methods for bond order.
```
xinglong-zhang: Implemented in other PR or not yet?

chemsmart/io/xtb/file.py
```diff
@@ -0,0 +1,1707 @@
...
+    the number of atoms.
+
+    Args:
+        filename (str): Path to the hessian file
+    """
+
+    def __init__(self, filename):
+        self.filename = filename
+
+    # TODO: Add parsing methods for hessian.
```
xinglong-zhang: Doesn't thermochemistry calcs require these? If these are the same as in the main.out file, then can we just read from there, if it's already parsed there.

chemsmart/io/xtb/input.py
```diff
@@ -0,0 +1,638 @@
+import os
+from functools import cached_property
+
+from chemsmart.utils.mixins import FileMixin
+
+
+class XTBInput(FileMixin):
```
xinglong-zhang, line 7: Does xtb job sometimes require not only the input but also the "side inputs" or auxiliary inputs too?

chemsmart/io/xtb/output.py
```diff
@@ -0,0 +1,481 @@
...
+
+    def __init__(self, folder):
+        """
+        Initialize XTBOutput with a folder path or XTBFolder instance.
+        """
+        self.folder = (
+            folder if isinstance(folder, XTBFolder) else XTBFolder(folder)
+        )
+
+    def __getattr__(self, name):
```
xinglong-zhang, line 46: is this only for individual files in the folder?

chemsmart/io/xtb/output.py
```diff
@@ -0,0 +1,481 @@
...
+
+    @property
+    def optimized_structure(self):
+        """
+        Get the final optimized structure (only if geometry optimization converged).
+
+        This method attempts to read the optimized structure from multiple sources:
+        1. First, try to read from optimized geometry file (xtbopt.xyz, xtbopt.sdf, etc.)
+
+        2. If xtbopt.* file is not available, fall back to the last structure
```
xinglong-zhang: if it is not available, it means opt has problem and we cannot simply ues the last structure - it may remain unoptimised.

chemsmart/io/xtb/route.py
```diff
@@ -0,0 +1,600 @@
...
+    XTB_ALL_METHODS,
+    XTB_ALL_OPT_LEVELS,
+    XTB_ALL_SOLVENT_IDS,
+    XTB_ALL_SOLVENT_MODELS,
+)
+
+logger = logging.getLogger(__name__)
+
+
+class XTBRoute:
```
xinglong-zhang, line 13: quite a lot of stuff here and much more complex than Gaussian/ORCA.

xinglong-zhang (changes requested):
@Huiwen-Tan, some comments for your consideration. We can merge this PR soon!

chemsmart/io/xtb/folder.py
Huiwen-Tan: fixed in commit https://github.com/xinglong-zhang/chemsmart/pull/588/changes/b7fd893c6b4b399326396003c5fa5bf2aaa67a23.

chemsmart/io/xtb/folder.py
Huiwen-Tan: Thanks for pointing that out! PDB is now supported as a parseable geometry format, and I've updated the corresponding parser logic in commit https://github.com/xinglong-zhang/chemsmart/pull/588/changes/f8097adc0705c1e4d57c01f778e5820eafbd9902.

chemsmart/cli/thermochemistry/thermochemistry.py
Huiwen-Tan: I have expanded the command docstring to clarify the directory restrictions in commit https://github.com/xinglong-zhang/chemsmart/pull/588/changes/aafe83554bbe5f49ca158e1d34fc6dae0514e48f.

For thermochemistry, directory discovery is intentionally non-recursive. A single xTB calculation at any depth can still be processed explicitly with -f path/to/main.out. Can refer to [this comment](https://github.com/xinglong-zhang/chemsmart/pull/588#issuecomment-5032380459) for the design concept. 

This behavior is already documented in the RTD under Thermochemistry Analysis → Batch processing of xTB calculation directories, including an example directory layout and the validation rules. I have now aligned the docstring with that RTD description as well.

chemsmart/io/xtb/folder.py
Huiwen-Tan: added in commit https://github.com/xinglong-zhang/chemsmart/pull/588/changes/e2a46eaa30f1590170fad905800c2cb94d4b91d8.

chemsmart/database/utils.py
Huiwen-Tan: Thanks for the suggestion. I took another look at the responsibilities of these utility modules.

`chemsmart.utils.datasets` mainly contains generic tabular/pKa dataset abstractions, whereas `chemsmart.database.utils` contains database-specific functionality such as SQLite connections and schema validation, record/trajectory IDs, serialization, and database query helpers. Merging them would therefore combine two largely unrelated concerns into a very large module. Since most of the latter utilities support the `chemsmart.database` functionality directly, I initially kept them colocated there.

I also kept `chemsmart.database` as a top-level component because it represents a persistence/data-access subsystem rather than a CHEMSMART job. In particular, `Database`, `DatabaseQuery`, `DatabaseInspector`, and `DatabaseExporter` do not inherit from `Job` or use a `JobRunner`; only assembly behaves somewhat like a processing task. Moving the whole component under `chemsmart.jobs` might therefore imply job-execution semantics that do not apply to most of it.

I think the overall package organization deserves a separate discussion. If we decide to reorganize it, I would prefer to do so in a dedicated follow-up PR, since it is outside the scope of the current `xtb_io` branch.

xinglong-zhang: Sounds logical. I was thinking if it needs to be a job like thermochemistry that may allow for submission (Lewen is implementing this) if database assembly becomes CPU intensive or if later we want to allow for parallelisation. But I think currently it may be okay. We can discuss.

chemsmart/io/xtb/file.py
Huiwen-Tan: You are right. I forgot to remove `xtb_version` after introducing the inherited version property. I have removed `xtb_version` in commit https://github.com/xinglong-zhang/chemsmart/pull/588/changes/77da537c0293cbe3cd27c5e768a827d8bde56ff0 and now use the inherited `version` property instead.

chemsmart/io/xtb/folder.py
Huiwen-Tan: Since PDB files are now supported for reading and can be used as geometry inputs, I noticed that the documentation at https://chemsmart.readthedocs.io/en/latest/molecule-input-formats.html may also need to be updated accordingly. I can prepare a separate PR to update the RTD documentation if that would be helpful.

xinglong-zhang:
Sure, can update RTD accordingly.
________________________________

chemsmart/io/xtb/file.py
Huiwen-Tan: fixed in commit https://github.com/xinglong-zhang/chemsmart/pull/588/changes/8fc47c97947e02b7017d009ecfa915663e44e690.

chemsmart/io/xtb/file.py
Huiwen-Tan: fixed in commit https://github.com/xinglong-zhang/chemsmart/pull/588/changes/8fc47c97947e02b7017d009ecfa915663e44e690.

chemsmart/io/xtb/file.py
Huiwen-Tan: fixed in commit https://github.com/xinglong-zhang/chemsmart/pull/588/changes/8fc47c97947e02b7017d009ecfa915663e44e690.

chemsmart/io/xtb/file.py
Huiwen-Tan: Thanks—I checked this. Frequencies, IR intensities, and related data do overlap with `Gaussian16Output`, but xTB's `g98.out` is only a truncated fake Gaussian output. It includes additional Raman Activ and Depolar lines (so the vibrational mode block starts at `i+7` rather than Gaussian's `i+5`) and has no normal termination. Feeding it to `Gaussian16Output` may recover the frequencies (which can already be parsed from the main xTB output), but the vibrational modes remain empty.

chemsmart/io/xtb/file.py
Huiwen-Tan: xTB does support open-shell radicals via `--uhf`. These properties return `None` because xTB does not print separate Alpha/Beta eigenvalue blocks like Gaussian; it prints one MO table with occupations. The number of unpaired electrons is already parsed from the SETUP section, and we already have an open-shell test (`p_benzyne` with `--uhf 2`).

chemsmart/io/xtb/output.py
Huiwen-Tan: `optimized_structure` is already guaranteed by normal termination and geometry-optimization convergence, so an unconverged opt returns `None` and is never treated as optimized.

A successful xTB optimization always writes an `xtbopt.*` file. The fallback to the last `xtbopt.log` frame is only for the case where that structure file exists but is not yet parseable by CHEMSMART (e.g. `xtbopt.coord`, `xtbopt.poscar`). In that case, `xtbopt.log` is still written as an XYZ trajectory, and for a converged optimization its last frame corresponds to the optimized geometry. I’ve clarified this in the docstring in commit https://github.com/xinglong-zhang/chemsmart/pull/588/changes/8edb2a1a7ffed0c01af3fc22499f338ed20d56bb.

chemsmart/io/xtb/input.py
Huiwen-Tan: Yes. Besides the geometry file, xTB can take optional detailed input via `--input` (xcontrol: $`constrain` / `$fix` / `$md` / …). `XTBInput` currently parses that detailed input file.

chemsmart/io/xtb/output.py
Huiwen-Tan: Yes. `__getattr__` only delegates attribute access to the individual file parsers in the calculation directory. Folder-level discovery remains on `XTBFolder`, while `XTBOutput` provides the public interface to the parsed results. For example, `output.homo_energy` is resolved via `XTBMainOut`, so callers don't need to access `output.main_out.homo_energy` explicitly.

chemsmart/io/xtb/file.py
Huiwen-Tan: Implemented in commit https://github.com/xinglong-zhang/chemsmart/pull/588/changes/b6ad7952285e1c0868ac90bb3cd1378e36de4ad7.

chemsmart/io/xtb/file.py
Huiwen-Tan: Implemented in commit https://github.com/xinglong-zhang/chemsmart/pull/588/changes/b6ad7952285e1c0868ac90bb3cd1378e36de4ad7.

chemsmart/io/xtb/file.py
Huiwen-Tan: This parser is for the full Cartesian Hessian matrix, which is not available in `main.out`. The thermochemistry quantities are parsed from `main.out`, but the underlying Hessian is only written to the standalone `hessian` file. This has now been implemented in commit https://github.com/xinglong-zhang/chemsmart/pull/588/changes/b6ad7952285e1c0868ac90bb3cd1378e36de4ad7.

xinglong-zhang (approved):

PR #589, opened 2026-07-02
Update config.py

Ayson-HAN:
Add GROMACS support to ChemSmart configuration so users can configure the GROMACS executable path via make configure / chemsmart config gromacs.

### DIFF
--- a/chemsmart/cli/config.py
+++ b/chemsmart/cli/config.py
@@ -312,6 +312,7 @@ class Config:
             ("Gaussian g16", "~/bin/g16", "Gaussian g16 folder"),
             ("ORCA", "~/bin/orca_6_0_0", "ORCA folder"),
             ("NCIPLOT", "~/bin/nciplot", "NCIPLOT folder"),
+            ("GROMACS", "~/bin/gromacs", "GROMACS folder"),
         ]:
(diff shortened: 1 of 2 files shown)


chemsmart/cli/config.py
```diff
@@ -312,6 +312,7 @@ def configure_paths_interactively(self) -> None:
             ("Gaussian g16", "~/bin/g16", "Gaussian g16 folder"),
             ("ORCA", "~/bin/orca_6_0_0", "ORCA folder"),
             ("NCIPLOT", "~/bin/nciplot", "NCIPLOT folder"),
+            ("GROMACS", "~/bin/gromacs", "GROMACS folder"),
```
liu-jing-yi, line 315: This will look for lines containing `~/bin/gromacs` in .yaml files when calling `update_yaml_files()`. So the current behaviour will not change if the .yaml files are not updated accordingly. Please add Gromacs sections in existing .yaml files also, e.g., `SLURM.yaml`, `PBS.yaml`, `server.yaml`, etc..

liu-jing-yi:
Please also add tests for this new configuration features in `test_YAMLIO.py`. @Ayson-HAN

PR #596, opened 2026-07-06
FIX: fix to thermochemistry

xinglong-zhang:
Fix to thermochemistry so that where Gaussian output file prints "****" for overlfowed values in Rotational temperature or Rotational Constants, the molecular thermochemistry can still be parsed.

Geometries close to linear will be treated as linear, instead of non-linear, as in Gaussian output file KOH.log.


chemsmart/analysis/thermochemistry.py
```diff
@@ -111,6 +111,7 @@ def __init__(
             i * (units._amu * (units.Ang / units.m) ** 2)
             for i in self.moments_of_inertia
         ]
+        print(f"Moments of inertia: {self.I}")
```
xinglong-zhang, line 114: to remove.

xinglong-zhang (commented):
to update

PR #602, opened 2026-07-07
Avoid dynamic attributes in GROMACS MDP writer

Ayson-HAN:
This PR addresses the review comment about avoiding dynamic attributes in the GROMACS MDP writer.

Changes:

- Remove `getattr` / dynamic attribute access from `GromacsInputWriter`
- Add explicit optional MDP-related attributes to `GromacsJob`
- Keep task-specific MDP defaults inside the writer instead of adding them to project YAML
- Use direct attribute access with a None fallback helper in the MDP writer
- Replace safe workflow `getattr(...)` calls in the runner with direct `job.workflow` access

Tests:

- `python -m py_compile chemsmart/jobs/gromacs/job.py`
- `python -m py_compile chemsmart/jobs/gromacs/runner.py`
- `python -m py_compile chemsmart/jobs/gromacs/writer.py`
- `python -m py_compile chemsmart/settings/gromacs.py`
- `python -m pytest tests/test_gromacs_runner.py tests/test_gromacs_settings.py -q`

Result:

- 45 passed

### DIFF
--- a/chemsmart/jobs/gromacs/job.py
+++ b/chemsmart/jobs/gromacs/job.py
@@ -1,7 +1,10 @@
 from __future__ import annotations
 
+"""
+GROMACS job definitions.
+"""
+
 from pathlib import Path
-from types import SimpleNamespace
 
 from chemsmart.jobs.job import Job
 
@@ -46,6 +49,13 @@ class GromacsJob(Job):
         barostat=None,
         constraints=None,
         constraint_algorithm=None,
+        nsteps=None,
+        emtol=None,
+        emstep=None,
+        tau_t=None,
+        tc_grps=None,
+        tau_p=None,
+        compressibility=None,
         box_type="cubic",
         box_distance=1.0,
         solvent_file=None,
@@ -92,14 +102,20 @@ class GromacsJob(Job):
             else None
         )
         self.boxed_structure_file = (
-            Path(boxed_structure_file) if boxed_structure_file else None
+            Path(boxed_structure_file)
+            if boxed_structure_file
+            else None
         )
         self.solvated_structure_file = (
-            Path(solvated_structure_file) if solvated_structure_file else None
+            Path(solvated_structure_file)
+            if solvated_structure_file
+            else None
         )
         self.ions_tpr_file = Path(ions_tpr_file) if ions_tpr_file else None
         self.ionized_structure_file = (
-            Path(ionized_structure_file) if ionized_structure_file else None
+            Path(ionized_structure_file)
+            if ionized_structure_file
+            else None
         )
 
         self.force_field = force_field
@@ -112,6 +128,14 @@ class GromacsJob(Job):
         self.barostat = barostat
         self.constraints = constraints
         self.constraint_algorithm = constraint_algorithm
+
+        self.nsteps = nsteps
+        self.emtol = emtol
+        self.emstep = emstep
+        self.tau_t = tau_t
+        self.tc_grps = tc_grps
+        self.tau_p = tau_p
+        self.compressibility = compressibility
 
         self.box_type = box_type
         self.box_distance = box_distance
@@ -180,7 +204,8 @@ class GromacsJob(Job):
         ]
 
         return all(
-            path is not None and Path(path).exists() for path in required_files
+            path is not None and Path(path).exists()
+            for path in required_files
         )
 
     def has_required_full_setup_inputs(self):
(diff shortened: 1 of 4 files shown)


chemsmart/jobs/gromacs/runner.py
```diff
@@ -646,18 +659,23 @@ def _validate_gromacs_inputs(self, job):
...
+                "Missing required GROMACS input files: "
+                + ", ".join(missing)
             )
 
     def _validate_full_setup_inputs(self, job):
         """
         Validate required inputs for full setup workflow.
         """
-        input_file = job.input_pdb or job.structure_file
+        input_file = getattr(job, "input_pdb", None) or getattr(
```
liu-jing-yi: We can keep using `job.input_pdb` here. Again, as we discussed, `getattr()` is for dynamic attributes. However, the current Gromacs implementation mostly relies on fixed names on objects that already define those attributes.  In this way, `getattr()` is not necessary, because It adds noise without protecting against real failure modes.

chemsmart/jobs/gromacs/runner.py
```diff
@@ -646,18 +659,23 @@ def _validate_gromacs_inputs(self, job):
...
+            job,
+            "structure_file",
+            None,
+        )
 
         required_files = {
             "input_pdb|structure_file": input_file,
-            "mdp_file": job.mdp_file,
+            "mdp_file": getattr(job, "mdp_file", None),
```
liu-jing-yi: Similar to the previous comment. Please also check elsewhere and make some changes.

chemsmart/jobs/gromacs/writer.py
```diff
@@ -142,11 +148,64 @@ def _build_nvt_mdp(self):
...
 
-    def _get(self, name, default=None):
+    def _build_npt_mdp(self):
+        """
+        Build a default NPT .mdp file.
+        """
+        temperature = self._value_or_default(self.job.temperature, 300)
+        pressure = self._value_or_default(self.job.pressure, 1.0)
+        timestep = self._value_or_default(self.job.timestep, 0.002)
+        nsteps = self._value_or_default(self.job.nsteps, 50000)
```
liu-jing-yi, line 158: Correct. This aligns with explicit over implicit design.

chemsmart/jobs/gromacs/job.py
```diff
@@ -207,17 +232,6 @@ def set_folder(self, folder):
         if self._use_default_tpr_file:
             self.tpr_file = Path(self.folder) / f"{self.label}.tpr"
 
-    def _output(self):
```
liu-jing-yi, line 210: Please do not remove this. Removal of _output lines will break existing tests.

liu-jing-yi (changes requested):

Ayson-HAN:
I updated this PR to keep task-specific MDP defaults inside the writer rather than adding them to `GromacsProjectSettings`.

I also confirmed that `settings/gromacs.py` no longer contains task-specific MDP parameters such as `emtol`, `emstep`, `tau_t`, `tau_p`, `compressibility`, or `nsteps`.

The MDP writer now uses explicit job attributes with `_value_or_default` instead of dynamic `getattr`.

chemsmart/jobs/gromacs/runner.py
```diff
@@ -71,12 +72,11 @@ def __init__(
...
 
     @property
-    @lru_cache(maxsize=12)
     def executable(self):
         """
-        Return the GROMACS executable configuration for the current server.
+        Return the configured GROMACS executable.
         """
-        return GromacsExecutable.from_servername(servername=self.server.name)
+        return self._get_executable()
```
liu-jing-yi: what `_get_executable()` returns is a string (e.g. "gmx"), but the base `JobRunner._update_os_environ()` expects an Executable object with an .env attribute (same pattern as Gaussian/ORCA runners). For this property, can keep it as it is in the original version.

chemsmart/jobs/gromacs/runner.py
```diff
@@ -139,10 +139,10 @@ def _get_executable(self):
         The direct runner option has the highest priority. If it is not given,
         fall back to GromacsExecutable.
         """
-        if self.gmx_executable is not None:
+        if getattr(self, "gmx_executable", None) is not None:
```
liu-jing-yi: `gmx_executable is declared` in `__init__`, so we'd better use direct attribute access instead of `getattr`. For this method, can keep it as the original version.

chemsmart/jobs/gromacs/runner.py
```diff
@@ -401,7 +401,7 @@ def _prepare_environment(self, env=None):
         Prepare environment variables for subprocess execution.
         """
         run_env = os.environ.copy()
-        run_env.update(self.gmx_env)
+        run_env.update(getattr(self, "gmx_env", {}) or {})
```
liu-jing-yi: Same as previous comment, we can use direct attributes most of the time (` run_env.update(self.gmx_env) ` here). Please change back to the original version and check elsewhere to keep consistency.

liu-jing-yi (changes requested):

Ayson-HAN:
Fixed. I restored direct access for gmx_modules, gmx_source_scripts, and fake as well, since these are fixed attributes initialized by the runner or its base class. I also rechecked the file and there are no remaining getattr(self, ...) usages in the GROMACS runner.

PR #605, opened 2026-07-08
Add the scratch configuration option

ZihanL1:


ZihanL1:
Add the scratch configuration option. Now users can customize scratch pathway when make configure

xinglong-zhang:
@ZihanL1, as discussed, please update the codes to run in job folder if scratch folder is not found, instead of currently raising an error that scratch_dir is not set.

xinglong-zhang (approved):

Issue #606, opened 2026-07-10
version 3.0.2 breaks for obtaining molecule object from pubchem ID


xinglong-zhang:
The following job has run well before

`chemsmart run --fake gaussian -p test --pubchem 222 -l ammonia -c 0 -m 1 opt`

however, now it raiases the following error:

> Traceback (most recent call last):
>   File "/Users/xinglongzhang/miniconda3/envs/chemsmart/bin/chemsmart", line 6, in <module>
>     sys.exit(main())
>   File "/Users/xinglongzhang/bin/chemsmart/chemsmart/cli/main.py", line 91, in main
>     entry_point(obj=obj)
>   File "/Users/xinglongzhang/miniconda3/envs/chemsmart/lib/python3.10/site-packages/click/core.py", line 1161, in __call__
>     return self.main(*args, **kwargs)
>   File "/Users/xinglongzhang/miniconda3/envs/chemsmart/lib/python3.10/site-packages/click/core.py", line 1082, in main
>     rv = self.invoke(ctx)
>   File "/Users/xinglongzhang/bin/chemsmart/chemsmart/utils/cli.py", line 109, in invoke
>     return super().invoke(ctx)
>   File "/Users/xinglongzhang/miniconda3/envs/chemsmart/lib/python3.10/site-packages/click/core.py", line 1697, in invoke
>     return _process_result(sub_ctx.command.invoke(sub_ctx))
>   File "/Users/xinglongzhang/miniconda3/envs/chemsmart/lib/python3.10/site-packages/click/core.py", line 1697, in invoke
>     return _process_result(sub_ctx.command.invoke(sub_ctx))
>   File "/Users/xinglongzhang/bin/chemsmart/chemsmart/utils/cli.py", line 109, in invoke
>     return super().invoke(ctx)
>   File "/Users/xinglongzhang/miniconda3/envs/chemsmart/lib/python3.10/site-packages/click/core.py", line 1697, in invoke
>     return _process_result(sub_ctx.command.invoke(sub_ctx))
>   File "/Users/xinglongzhang/bin/chemsmart/chemsmart/utils/cli.py", line 109, in invoke
>     return super().invoke(ctx)
>   File "/Users/xinglongzhang/miniconda3/envs/chemsmart/lib/python3.10/site-packages/click/core.py", line 1675, in invoke
>     rv = super().invoke(ctx)
[...]

PR #607, opened 2026-07-10
Fix: running Gaussian using Pubchem ID

liu-jing-yi:
Issue #606 occurs because the PubChem block is skipped when not using `-f`. Moving it out of the `if filename` loop should fix this problem. 
The `not is_pka_table_input` condition was added to avoid unnecessary PubChem call in edge case such as `-f pka_table.csv --pubchem 222 pka batch`.


xinglong-zhang (changes requested):
Could you add a test for this CLI command?

liu-jing-yi:
> Could you add a test for this CLI command?

Test added.

xinglong-zhang:
tests are failing - linting needed?

xinglong-zhang (approved):

PR #610, opened 2026-07-10
feat: add steric and sterimol property in Molecule class

ryanleung521:
- Added 2D and 3D steric descriptors properties in Molecule class by using rdkit. 
- Added sterimol parameters properties in Molecule class by using the new dbstep package. 
- Tested the newly added properties


.vscode/settings.json
```diff
@@ -0,0 +1,9 @@
+{
```
xinglong-zhang, line 1: please add .vscode to .gitignore so that all .vscode changes will be ignored.

chemsmart/io/molecules/structure.py
```diff
@@ -891,27 +893,86 @@ def rotational_temperatures(self):
...
+        return [
+            units._hplanck**2 / (8 * np.pi**2 * moi_in_SI_units[i] * units._k)
+            for i in range(3)
+        ]
+
+    @property
+    def sterimol_parameters(self):
+        """dbstep.dbstep | None: Holds the most recently computed DBSTEP results object,
+        or None if no compute_sterimol_parameters() has been run yet."""
+        return getattr(self, "_sterimol_parameters", None)
```
xinglong-zhang: it's better not to use getattr() method, since if the molecule object is already created/instantiated, then the properties that it has should also been created at the same time.

chemsmart/io/molecules/structure.py
```diff
@@ -891,27 +893,86 @@ def rotational_temperatures(self):
...
+        Args:
+            atom1 (int): The index of the anchor atom (1-based depending on framework).
+            atom2 (int): The index of the directional atom establishing the measurement axis.
+            grid_measure (bool): Toggle between classical analytical or cubic grid-based measurements.
+
+        Returns:
+            dbstep.dbstep: A DBSTEP object containing the computed L, Bmin, and Bmax parameters.
+        """
+
+        import dbstep.Dbstep as db
```
xinglong-zhang: instead of importing dbstep python module, can we calculate these parameters directly?

xinglong-zhang (changes requested):
Thanks @ryanleung521, things are looking good. I am just wondering if we can calculate these steric paramters from the geometry instead of from dbstep dependency?

ryanleung521:
I had been struggling with the computation accuracy of the sterimol calculation results for a very long time. I figured out it was the rdkit H atom vdw radius issue, but on top of that there is still different factors contributing to this error. If it is possible please take a look.

ryanleung521:
Eventually verified the algo....

PR #616, opened 2026-07-18
Standard name

xinglong-zhang:
Standardize program name as `CHEMSMART` and command as `chemsmart` - leave `CHANGELOG.md` untouched.


xinglong-zhang (commented):
reviewed

PR #617, opened 2026-07-18
Iterate v2

Lewen-WangCC:
Summary
This PR redesigns the CHEMSMART Iterate workflow to support true multi-site, multi-substituent structure generation.
Compared with the implementation on main, this version:
replaces flat single-substitution enumeration with group- and slot-based combination generation;
upgrades the original single-substituent Lagrange multiplier method to Joint Lagrange Geometry Optimization (JLGO);
introduces a fused JLGO strategy that reduces unnecessary SLSQP calculations while improving structural robustness;
integrates RDKit ETKDGv3 as a complete alternative analyzer and makes ETKDG the default algorithm;
unifies both algorithms under the same multi-substituent analyzer and configuration pipeline.
Combination generation
The main implementation creates only flat combinations of:
one skeleton × one skeleton link position × one substituent
Each task therefore contains exactly one substituent assignment, even when a skeleton defines several attachment sites.
This PR replaces that model with an IterateCombination containing one or more explicit assignments. YAML configurations can associate substituents with numbered groups and define skeleton positions through explicit slots.
For every attachment position, Iterate constructs the choices:
keep the original branch / substituent A / substituent B / ...
It then builds the required Cartesian products and excludes the all-empty result. This allows one generated structure to contain two, three, or more substituents at different skeleton positions.
Two expansion modes are supported:
independent: expands each explicit slot independently, while positions within the same slot are still combined;
global: merges all explicit slots and performs one Cartesian product across every participating position.
The original link_index shorthand remains available for independent single-site screening. Explicit slots are used when multiple positions must participate in the same combination.
Additional improvements include:
substituents may be assigned to one or more groups;
all attachment sites in a combination are preprocessed in one batch;
skeleton atom indices are remapped only once after branch removal;
combination labels are deterministic and include every assignment;
duplicate labels, overlapping attachment positions and inconsistent groups are rejected during validation;
molecule objects remain in shared pools and are copied only when a worker executes a combination.
From Lagrange multipliers to JLGO
The implementation on main optimizes one rigid substituent at a time using a six-dimensional SLSQP problem:
3 linking-atom coordinates + 3 Euler angles
It samples 96 sphere directions and six axial rotations by default, potentially running up to 576 SLSQP optimizations for one substituent. Because different substituents are handled independently, the optimizer cannot directly account for steric interactions between substituents in the same final structure.
JLGO replaces this with one joint 6K-dimensional optimization for K substituents. Each substituent contributes its linking-atom coordinates and Euler angles, while all substituents are optimized simultaneously.
The joint problem includes:
one covalent bond-sphere equality constraint per substituent;
substituent–skeleton VDW constraints;
substituent–substituent VDW constraints;
a smooth soft-minimum non-bonded-distance objective;
a calibrated WCA repulsive contribution;
RDKit covalent and van der Waals radii.
Single substitution (K = 1) and multi-substitution (K > 1) use the same JLGO code path. The analyzer returns either the complete combined structure or no structure; it never returns a partially inserted combination.
The public CLI algorithm name is now jlgo. Legacy Lagrange names remain registry aliases for configuration compatibility but are no longer exposed as CLI algorithm subcommands.
Fused strategy and computational efficiency
The JLGO default is a fused strategy selected from the standalone Test 7.2/Test 8 comparisons. It combines several complementary stages rather than sending every sampled geometry directly to SLSQP.
The start-generation pipeline is:
link-sphere and orientation sampling
→ skeleton-region exclusion
→ diverse per-substituent candidate pools
→ incremental substituent–substituent feasibility pruning
→ bounded beam search
→ top-score preselection
→ greedy diversity selection
→ SLSQP
Key efficiency improvements include:
Region exclusion removes linking positions and rigid poses that already collide with the skeleton.
Per-substituent candidate pools retain only high-clearance, geometrically diverse poses instead of optimizing every raw sample.
Feasible-domain pruning rejects substituent–substituent collisions as each substituent is added to a partial joint combination.
Bounded beam search limits the theoretical M^K growth of joint candidate enumeration.
Preselection and greedy diversity selection avoid spending SLSQP starts on nearly identical geometries.
Adaptive sampling first uses a smaller coarse sampling level and expands to the full level only when the coarse stage cannot produce an acceptable structure.
Convergence detection terminates an SLSQP start when the objective has plateaued and the constraints are already satisfied.
Early acceptance stops the remaining starts once an acceptable structure is found.
With the default configuration, the final candidate set is reduced to at most eight selected SLSQP starts per sampling level, instead of directly applying SLSQP to hundreds or exponentially many raw geometries.
Robustness improvements
JLGO adds several recovery and validation layers:
every candidate is checked against explicit equality and inequality tolerances;
a constraint-feasible solution may be accepted even when SciPy does not set success=True;
geometrically diverse starts reduce sensitivity to local minima;
the adaptive full stage is attempted when the coarse stage fails;
a structural quality gate checks short non-bonded contacts, minimum distances and aggregate VDW overlap;
the numerically feasible candidate closest to the quality boundary can be re-optimized with the quality requirements added as explicit constraints;
if callback-based convergence detection produces no acceptable result, ordinary SLSQP is retried on the same starts;
invalid indices, missing atomic radii and invalid numerical configuration values are rejected before optimization;
failures are isolated to the affected combination and do not produce partially assembled molecules.
At the runner level, preprocessing, algorithm failure, timeout and worker-crash outcomes are recorded separately. Successful structures from partially failed runs are retained and documented in the Iterate report.
ETKDG integration
This PR adds a dedicated RDKit ETKDGv3 analyzer and registers it alongside JLGO. ETKDG is now the built-in default because it provides a mature and generally robust structure-generation path.
For a multi-substituent combination, ETKDG:
batch-preprocesses all skeleton attachment positions;
converts the skeleton and every substituent into separate RDKit fragments;
combines all fragments into one RDKit molecule;
adds every skeleton–substituent bond explicitly;
embeds the complete assembly once.
This avoids sequential insertion and repeated molecule conversions that could lose connection topology or produce fragmented structures.
Two embedding modes are provided:
local mode (default): preserves the processed skeleton through an RDKit coordinate map and restores its coordinates exactly after embedding and optional force-field refinement;
global mode: re-embeds the complete combined molecule.
ETKDG also supports:
multiple conformers with lowest-UFF-energy selection;
reproducible generation through a validated random seed;
configurable embedding iterations, random-coordinate initialization and chirality enforcement;
optional UFF, MMFF94 or MMFF94s refinement;
graceful fallback to the successfully embedded conformer when force-field refinement is unavailable;
preservation of atom order, charge, multiplicity and frozen-atom metadata.
Supporting workflow changes
YAML replaces the original TOML Iterate configuration workflow.
Algorithm options are validated through a shared registry.
Configuration priority is:
built-in defaults < YAML algorithm block < explicit CLI options.
Normal terminal output uses a progress bar and compact totals, while --debug restores detailed worker and RDKit diagnostics.
Every executed run produces a Gaussian-style plain-text Iterate report with per-combination results, output paths and error codes.
The Iterate tests now use small molecular fixtures and manually generated golden XYZ files for JLGO and ETKDG, together with focused coverage for combination traversal, expansion modes, configuration errors, timeouts, partial success and separate outputs.


xinglong-zhang:
Conflicts to be resolved, @Lewen-WangCC

Issue #613, opened 2026-07-15
CI workflow failed for tests_win (3.10, windows-latest)


liu-jing-yi:
CI workflow failed for Windows with auto-updated condo version (@v4), but passed for Mac and Linux. This may require further testing on a virtual machine.
```
Node 20 is being deprecated. This workflow is running with Node 24 by default. If you need to temporarily use Node 20, you can set the ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION=true environment variable. For more information see: https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/
Run conda-incubator/setup-miniconda@v3
Gathering Inputs...
Creating bootstrap condarc file in C:\Users\runneradmin\.condarc...
Ensuring installer...
Setup environment variables...
Parsing environment...
Applying initial configuration...
Initializing conda shell integration...
Adding tools to 'base' env...
Ensuring environment...
Updating 'chemsmart' env from conda env update...

Error: The process 'C:\Miniconda\condabin\conda.bat' failed with exit code 1
```

xinglong-zhang:
@liu-jing-yi, looks like your solution is correct, but i don't know if using latest conda version on windows, will the installation work? Can you test this please?

liu-jing-yi:
Addressed in PR #620 .

xinglong-zhang:
Fix in PR https://github.com/xinglong-zhang/chemsmart/pull/622 to use miniforge and libmamba as solvents for GitHub Actions.

PR #620, opened 2026-07-18
Fix: turn off conda auto-update for Windows CI workflow

liu-jing-yi:
Issue #613 occurs because updating Conda on Windows rewrites `conda.bat` while the script is actively executing. On Linux and macOS, this isn't an issue because the shell handles active scripts differently, preventing the file corruption that happens when packages are replaced on Windows. More details are available in [Conda Issue #14322](https://github.com/conda/conda/issues/14322).

To work around this and use the latest version of Conda on Windows, we can install a fresh instance of the latest Miniconda and disable auto-updates. A `Show conda version` step has been added to the workflow to confirm the latest Miniconda is successfully installed, as verified in this [CI run log](https://github.com/liu-jing-yi/chemsmart/actions/runs/29649265599/job/88092996011):
image

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -283,7 +283,8 @@ jobs:
           python-version: ${{ matrix.python-version }}
           environment-file: environment.yml
           activate-environment: chemsmart
-          channels: conda-forge,defaults
+          channels: defaults,conda-forge
+          channel-priority: strict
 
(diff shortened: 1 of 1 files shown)


xinglong-zhang:
Fix in PR https://github.com/xinglong-zhang/chemsmart/pull/622

PR #621, opened 2026-07-20
fix: case-sensitive .cdx file detection

liu-jing-yi:
- Fix: case-sensitive `.cdx` file detection.
- Test: add `test_molecule_from_filepath_cdx` and `test_molecule_from_filepath_cdx_uppercase_extension`.


xinglong-zhang:
@liu-jing-yi, I think parsing .cdxml works but parsing .cdx needs to be fixed.

xinglong-zhang (approved):

xinglong-zhang (approved):

PR #488, opened 2026-02-26
Gaussian pKa Implementation

liu-jing-yi:
Implement Gaussian pKa support around a isodesmic cycle that merges  thermochemistry module with solvent single points and ensures every HA/A⁻/HB/B⁻ job inherits the “Gaussian pKa calculation job” title, labels, and consistent levels of theory.

Extends chemsmart run gaussian pka with explicit proton-index, reference-acid, thermochemistry, and log-parsing flags so users can either launch fresh HA/A⁻/HB/B⁻ workflows or feed in completed outputs plus an experimental reference pKa; docs/tests now describe and cover the workflow end to end.

Example calling command:

For job submission:

1. single job submission: `chemsmart run gaussian -p my_project -f acid.xyz -c 0 -m 1 pka -pi 10 -t "proton exchange"  -r water.xyz  -rpi 1 -rc 0 -rm 1 -rp 14.0`
2. batch job submission: `chemsmart run gaussian -p my_project -f acid.xyz pka -o <acids_input_table>.csv`

For output analysis:

1. single specie analysis: `chemsmart run pka -ha acid_opt.log -a conjugate_base_opt.log -hb reference_acid_opt.log -b reference_base_opt.log -has acid_sp_smd.log -as conjugate_base_sp_smd.log -hbs reference_acid_sp_smd.log -bs reference_base_sp_smd.log -rp 6.75 -T 298.15`
2. batch analysis: `chemsmart run pka -O <acid_output_logfiles>.csv`

Example output:
```
==============================================================================
pKa Calculation - Dual-level Proton Exchange Scheme
==============================================================================
Reaction: HA + B⁻ → A⁻ + HB
Temperature: 373.15 K

Method:
  G_corr = qh-G(T) - E_gas  (from gas-phase freq calculation)
  G_soln = E_solv + G_corr  (solution free energy)
  ΔG_soln = [G(A⁻)_soln + G(HB)_soln] - [G(HA)_soln + G(B⁻)_soln]
  pKa = pKa_ref + ΔG_soln / (RT × ln10)
------------------------------------------------------------------------------

Gas-Phase Electronic Energies (E_gas, au):
  HA:  -345.7419436500
  A⁻:  -344.9153986020
  HB:  -365.8436493070
  B⁻:  -365.4561783660

Thermal Corrections (G_corr = qh-G - E_gas, au):
  HA:  0.0931931305
  A⁻:  0.0758935969
  HB:  0.1404528844
  B⁻:  0.1267467582

[...]

PR #623, opened 2026-07-20
Scratch fix

liu-jing-yi:
Fix scratch running bug triggered by merging PR #488:

In chemsmart/cli/run.py:
```
if scratch is None:
    scratch = False
```
accidentally blocks program defaults. These lines are removed in this PR.

In chemsmart/jobs/runner.py: 
the `type(self) is not JobRunner:` condition is added to prevent resolving scratch too early. Without this condition, `_set_scratch()` will try to figure out the scratch path for `JobRunner` instead of `GaussianJobRunner` or `ORCAJobRunner` (`type(ORCAJobRunner)` is not a `JobRunner`), which returns `scratch=False`. Consequently, when we set `SCRATCH_DIR: Null` for server and `SCRATCH: True` for specific programs, running `chemsmart sub --scratch gaussian ...` ends up with `scratch=False` and runs in job folder.


chemsmart/cli/run.py
```diff
@@ -56,23 +57,27 @@ def run(
     create_logger(debug=debug, stream=stream)
     logger.info("Entering main program")
 
-    # Local execution does not use scratch unless explicitly requested.
-    if scratch is None:
-        scratch = False
+    # True if user passed --scratch/--no-scratch (vs omit).
```
xinglong-zhang: can this just be `scratch`?

chemsmart/jobs/runner.py
```diff
@@ -167,13 +194,14 @@ class JobRunner(RegistryMixin):
...
         server,
-        scratch=None,
+        scratch=False,
         scratch_dir=None,  # Explicit scratch directory
         delete_scratch=False,
         fake=False,
         num_cores=None,
         num_gpus=None,
         mem_gb=None,
+        scratch_from_cli=False,
```
xinglong-zhang: `scratch_from_cli` should not be needed as a separate parameter, the value read in should be passed to `scratch` directly and behavior modified accordingly.

xinglong-zhang (approved):
try to not introduce additional variable scratch_from_cli, since this value can be passed to scratch directly.

xinglong-zhang:
```

User input | scratch | scratch_from_cli | Meaning
-- | -- | -- | --
(omit flag) | False | False | Use program default
--scratch | True | True | Force scratch on
--no-scratch | False | True | Force scratch off
```



this is confusing, you have False and True but scratch is off.

xinglong-zhang (changes requested):
wrongly pressed approve earlier, should be request changes.

chemsmart/jobs/runner.py
liu-jing-yi: Removed. Now the scratch behavior is governed solely by the `--scratch/--no-scratch` CLI flag, and omitting the flag defaults to` scratch=False` (ignoring any `SCRATCH: True` entry in `.yaml` files).

liu-jing-yi:
> ```
> 
> User input | scratch | scratch_from_cli | Meaning
> -- | -- | -- | --
> (omit flag) | False | False | Use program default
> --scratch | True | True | Force scratch on
> --no-scratch | False | True | Force scratch off
> ```
> 
> this is confusing, you have False and True but scratch is off.

This is replaced by an updated table in PR description.

liu-jing-yi:
> try to not introduce additional variable scratch_from_cli, since this value can be passed to scratch directly.

Removed.

chemsmart/jobs/gaussian/runner.py
```diff
@@ -112,18 +105,12 @@ def __init__(
 
         Args:
             server: Server configuration object for job execution.
-            scratch (bool, optional): Whether to use scratch directories.
```
xinglong-zhang, line 115: this small `scratch` should be preserved, since it is an argument in `__init__`.

chemsmart/jobs/gaussian/runner.py
```diff
@@ -112,18 +105,12 @@ def __init__(
...
             fake (bool): Whether to run in fake/test mode.
             scratch_dir (str, optional): Custom scratch directory path.
             **kwargs: Additional arguments passed to parent JobRunner.
         """
-        # Use default SCRATCH if scratch is not explicitly set
-        if scratch is None:
-            scratch = self.SCRATCH
         super().__init__(
             server=server,
-            scratch=scratch,
```
xinglong-zhang, line 126: small scratch is still used, in contrast to capital SCRATCH

chemsmart/cli/jobrunner.py
```diff
@@ -42,9 +42,9 @@ def click_jobrunner_options(f):
     )
     @click.option(
         "--scratch/--no-scratch",
-        default=None,
+        default=False,
         type=bool,
-        help="Run in scratch mode or without a scratch folder.",
+        help="Force scratch (--scratch) or job-folder (--no-scratch) mode. ",
```
xinglong-zhang: should not say "Force"?

chemsmart/jobs/gaussian/runner.py
xinglong-zhang: wait, i see your new design removes small scratch as an argument to all runners.

chemsmart/jobs/runner.py
```diff
@@ -188,11 +188,14 @@ def __init__(
...
-        self.scratch = scratch
+        self.scratch = bool(scratch)
         self._scratch_dir = scratch_dir  # Store user-defined scratch_dir
         self.delete_scratch = delete_scratch
 
-        if self.scratch:
+        # CLI builds a placeholder JobRunner before the typed runner exists.
+        # Resolving scratch here has no executable and would set scratch=False,
+        # defeating --scratch. Typed runners resolve in their __init__.
+        if self.scratch and type(self) is not JobRunner:
```
xinglong-zhang, line 239: This condition `type(self) is not JobRunner` will always be False and the `set_scratch()` cannot be executed, even if `self.scratch=True`?

chemsmart/jobs/runner.py
```diff
@@ -263,12 +266,14 @@ def _set_scratch(self):
             )
             self.scratch = False
         else:
-            # check that the scratch folder exists
             scratch_dir = os.path.expanduser(scratch_dir)
-            if not os.path.exists(scratch_dir):
-                raise FileNotFoundError(
-                    f"Specified scratch dir does not exist: {scratch_dir}"
+            if not os.path.isdir(scratch_dir):
```
xinglong-zhang: good check.

chemsmart/settings/templates/.chemsmart/server/local.yaml
```diff
@@ -9,14 +9,13 @@ SERVER:
...
     ##PROJECT: 13002374
-    SCRATCH: Null
+    SCRATCH_DIR: Null
     USE_HOSTS: True
     EXTRA_COMMANDS: |
         #extra commands to activate chemsmart environment in submission script
 GAUSSIAN:
     EXEFOLDER: ~/bin/g16 
     LOCAL_RUN: True  # local run True to use serial_command
-    SCRATCH: True  # set scratch to True to run in scratch folder
```
xinglong-zhang, line 19: Removing this will not allow users to toggle between run in scratch or no-scratch for program specific scratch specifications in server.yaml isn't it?

xinglong-zhang:
is this fixed now, can the scratch be turned on/off from server yaml under each program? have you tested runs on cluster? @liu-jing-yi

liu-jing-yi:
Reverted to the original design for scratch, please have a check @xinglong-zhang .

chemsmart/jobs/runner.py
xinglong-zhang: why is there this check needed? shouldn't the child class be checked instead of this parent JobRunner class?

tests/test_jobrunner.py
```diff
@@ -149,3 +154,134 @@ def test_fake_orca_does_not_duplicate_fake_suffix(
         assert Path(runner.job_gbwfile).name == "orca_opt_fake.gbw"
         assert Path(runner.job_errfile).name == "orca_opt_fake.err"
         assert Path(runner.job_outputfile).name == "orca_opt_fake.out"
+
+
+def _write_gaussian_project(tmp_path):
```
xinglong-zhang: can you move this to conftest.py as a pytest.fixture() where it can be reused? try not to define private functions inside test files.

xinglong-zhang (changes requested):
@liu-jing-yi

chemsmart/jobs/runner.py
liu-jing-yi: The design logic is that we skip scratch path setup unless this is a concrete runner. We can also move `_set_scratch()` into each subclass, but that duplicates the same logic in Gaussian, ORCA, NCIPLOT, etc.

tests/test_jobrunner.py
liu-jing-yi: Thanks for pointing this out. This is moved to `conftest.py` as a `pytest.fixture()`.

chemsmart/jobs/gaussian/runner.py
liu-jing-yi: This is an outdated version hence is marked as resolved.

chemsmart/cli/jobrunner.py
liu-jing-yi: This is an outdated version hence is marked as resolved. Same as other outdated comments.

chemsmart/jobs/runner.py
liu-jing-yi: Outdated version hence is marked as resolved. Same as other outdated comments.

chemsmart/settings/templates/.chemsmart/server/local.yaml
liu-jing-yi: This is an outdated version hence is marked as resolved. Same as other outdated comments.

xinglong-zhang (approved):

xinglong-zhang (approved):

PR #630, opened 2026-07-22
xTB Job Submission (SP/OPT/HESS)

Huiwen-Tan:


chemsmart/cli/utils.py
```diff
@@ -0,0 +1,40 @@
+import logging
```
xinglong-zhang, line 1: should this be placed in chemsmart.cli.utils.py

chemsmart/cli/xtb/opt.py
```diff
@@ -0,0 +1,51 @@
...
+from chemsmart.cli.xtb.xtb import xtb
+from chemsmart.utils.cli import MyCommand
+from chemsmart.utils.utils import check_charge_and_multiplicity
+
+logger = logging.getLogger(__name__)
+
+
+@xtb.command("opt", cls=MyCommand)
+@click_job_options
+@click.option(
```
xinglong-zhang: i imagine this optimization level option will apply not only to opt job but all other xtb jobs? it will be better to define a reusable decorator

chemsmart/io/xtb/file.py
```diff
@@ -213,6 +213,8 @@ def spin(self):
     def optimization_level(self):
         """Optimization level."""
         opt_level = self._get_setup_information("optimization level")
+        if opt_level and opt_level.lower() == "verytight":
```
xinglong-zhang, line 216: i think both "vtight" and "verytight" will work in xtb? in that case, this change of keyword is not needed.

chemsmart/jobs/xtb/runner.py
```diff
@@ -0,0 +1,254 @@
...
+        logger.debug(f"Generated xTB command: {command}")
+        return command
+
+    def _settings_args(self, settings):
+        # Keep command rendering centralized so run/sub/fake paths cannot drift
+        # when new xTB flags or jobtypes are added.
+        args = []
+        args.extend(self._gfn_args(settings.gfn_version))
+
+        if settings.jobtype == "opt":
```
xinglong-zhang, line 124: there is one jobtype "ohess" which is running opt followed by freq, can this be added? herein it's running opt and hess as two separate jobs.

chemsmart/jobs/xtb/runner.py
```diff
@@ -0,0 +1,254 @@
...
+        job.molecule.write_xyz(self.job_xyzfile, mode="w")
+
+    def _get_command(self, job):
+        executable = self.executable.get_executable()
+        command = [executable, self.job_xyzfile]
+        command.extend(self._settings_args(job.settings))
+        logger.debug(f"Generated xTB command: {command}")
+        return command
+
+    def _settings_args(self, settings):
```
xinglong-zhang: should this be `get_settings_args()` and as a public method.

chemsmart/jobs/xtb/runner.py
```diff
@@ -0,0 +1,254 @@
...
+        self.job_errfile = os.path.abspath(job.errfile)
+
+    def _write_input(self, job):
+        logger.info(f"Writing xTB geometry input file to: {self.job_xyzfile}")
+        job.molecule.write_xyz(self.job_xyzfile, mode="w")
+
+    def _get_command(self, job):
+        executable = self.executable.get_executable()
+        command = [executable, self.job_xyzfile]
+        command.extend(self._settings_args(job.settings))
```
xinglong-zhang: this reconstruction of commands from CLI actually allows this to be a job that can be submitted to a compute node when needed? Lewen was trying to implement this for other jobs, such as thermochemistry and this should be designed as an OOP class, so it can be used in any runner.

chemsmart/jobs/xtb/runner.py
```diff
@@ -0,0 +1,254 @@
...
+    """Fake xTB runner for CLI, submit, and parser smoke tests."""
+
+    FAKE = True
+
+    def __init__(self, server, scratch=None, fake=True, **kwargs):
+        super().__init__(server=server, scratch=scratch, fake=fake, **kwargs)
+
+    @property
+    @lru_cache(maxsize=12)
+    def executable(self):
```
xinglong-zhang, line 204: fake xTB will not have any executable

chemsmart/settings/templates/.chemsmart/server/local.yaml
```diff
@@ -45,6 +45,18 @@ ORCA:
     ENVARS: |
         export SCRATCH=~/scratch  # required if scratch is true
 
+XTB:
+    EXEFOLDER: null  # use xtb from activated conda environment/PATH
```
xinglong-zhang, line 49: is this `xtb` executable auto detected by the code?

chemsmart/settings/templates/.chemsmart/xtb/test.yaml
```diff
@@ -0,0 +1,18 @@
+sp:
```
xinglong-zhang, line 1: cool, so the design of xtb project settings is each job type having a setting?

chemsmart/settings/templates/.chemsmart/server/local.yaml
xinglong-zhang: saw this definition in XTBExecutable class!

xinglong-zhang: if this is defined, e.g., EXEFOLDER pointing to g-xtb, such that the command call is g-xtb, will that be possible?

xinglong-zhang: i saw that in XTBExecutable, this will be returned first if it is None, else it will return "xtb" keyword pointing to xtb call.

environment.yml
```diff
@@ -7,6 +7,7 @@ dependencies:
   - pymol-open-source=3.1.0
   - ffmpeg=7.1.1  # movie support for pymol
   - openbabel          # includes Python bindings on conda-forge
+  - xtb                # provides the xtb command-line executable via conda-forge
```
xinglong-zhang, line 10: nice

xinglong-zhang (changes requested):
@Huiwen-Tan, looking sharp here! some comments for consideration.

xinglong-zhang:
@Huiwen-Tan, is this ready to be reviewed and merged? Can you improve the code coverage by including neccessary tests?

chemsmart/settings/templates/.chemsmart/xtb/test.yaml
Huiwen-Tan: yes

chemsmart/cli/utils.py
Huiwen-Tan: sure, moved it in commit [a77d92f](https://github.com/xinglong-zhang/chemsmart/pull/630/commits/a77d92ff2e13dcfda514889a4ca27c22a75c9155).

chemsmart/io/xtb/file.py
Huiwen-Tan: That's right—both `vtight` and `verytight` are accepted by xTB. The normalization is intentional for CHEMSMART rather than xTB itself—we convert equivalent keywords to a canonical form before storing them so that downstream processing and database records use a consistent value.

chemsmart/cli/xtb/opt.py
Huiwen-Tan: fixed in commit https://github.com/xinglong-zhang/chemsmart/pull/630/changes/8d366bcd291c50d4df25b49acbadca3208eabeb4.

chemsmart/jobs/xtb/runner.py
Huiwen-Tan: fixed in commit https://github.com/xinglong-zhang/chemsmart/pull/630/changes/8d366bcd291c50d4df25b49acbadca3208eabeb4.

chemsmart/jobs/xtb/runner.py
Huiwen-Tan: Yes — rebuilding the command from the job settings allows the same job definition to be used for both local `run` and `sub` on a compute node (write XYZ → build the command → execute under the configured server/conda environment). 

After this PR is merged, Lewen and I will follow up on generalizing this into a reusable OOP design that can be shared across other runners.

chemsmart/jobs/xtb/runner.py
Huiwen-Tan: `FakeXTBJobRunner.executable` returns a stub `XTBExecutable(executable_folder=None, local_run=True)` only so `_get_command` can render the same command as a real run. `FakeXTBJobRunner.run` never invokes `Popen`, it writes a fake `.out` via `FakeXTB` for parser/CLI smoke tests. So there is no real executable dependency in fake mode.

chemsmart/jobs/xtb/runner.py
Huiwen-Tan: Thanks. I added support for xTB `--ohess` in commit https://github.com/xinglong-zhang/chemsmart/pull/630/changes/b43cc521c25efbd104dc5255da60be72ba90cee3 through the existing `opt + freq` abstraction rather than introducing a new `ohess` jobtype. When `jobtype="opt"` and `freq=True` (default), the runner generates `--ohess`; otherwise it generates `--opt`. This keeps the workflow consistent with Gaussian/ORCA settings.

chemsmart/settings/templates/.chemsmart/server/local.yaml
Huiwen-Tan: Good point. I'll address this in a separate PR for g-xTB support. Since g-xTB requires a modified xTB executable and is enabled through the `--gxtb` flag, I will add the corresponding executable handling, CLI/job setting, tests, and documentation.

chemsmart/io/xtb/route.py
```diff
@@ -193,7 +194,7 @@ def get_jobtype(self):
         Extract job type from route specification.
 
         Checks for various job type keywords in order of precedence:
-        1. Optimization: --opt, --omd, --metaopt, --ohess, --bhess
```
xinglong-zhang, line 196: what do --omd and --bhess do that they are now removed?

chemsmart/jobs/xtb/runner.py
```diff
@@ -0,0 +1,274 @@
...
+            logger.error(
+                f"No server file {self.server} is found for xTB: {exc}\n"
+                f"Available servers: {XTBExecutable.available_servers}"
+            )
+            raise
+
+    def _prerun(self, job):
+        self._assign_variables(job)
+
+    def _assign_variables(self, job):
```
xinglong-zhang, line 62: these are always repeated, when i set up it was for gaussian/orca, perhaps it is best to refactor as a jobrunner mixin in future..

chemsmart/jobs/xtb/writer.py
```diff
@@ -0,0 +1,113 @@
...
+            self._write_all(f)
+
+        logger.info(
+            f"Wrote xTB input file with "
+            f"{len(self.settings.constraints)} constraint(s)"
+        )
+
+    def _write_all(self, f):
+        """Write all currently supported xcontrol sections."""
+        self.write_constraints(f)
```
xinglong-zhang, line 60: might raise error if constraint is None? thus need to have a conditional check before this

xinglong-zhang (changes requested):
@Huiwen-Tan, things are looking great now, just a couple of comments for further consideration and we can merge.

Please check RTD very carefully, as I did not check.
Please also test this branch on local machine and cluster to make sure that the Gaussian, orca, nciplot and others are not affected and that this xtb works well as intended.

xinglong-zhang:
@Huiwen-Tan

chemsmart/io/xtb/route.py
Huiwen-Tan: `--omd` performs an optimization followed by a molecular dynamics simulation, while `--bhess` performs an optimization followed by a biased numerical Hessian calculation. They were removed because CHEMSMART does not currently support these xTB run types. Keeping them in the parser could incorrectly classify unsupported commands as optimization or single-point jobs. If we add support for them in the future, they should be introduced as dedicated job types rather than being classified as optimization jobs.

chemsmart/jobs/xtb/writer.py
Huiwen-Tan: Through the normal `write()` path this was already guarded by:
```
if self.settings.constraints:
    self._write_inp(folder)
```
so `_write_all()` was only reached when constraints were present. I've added an early return in `write_constraints()` in commit https://github.com/xinglong-zhang/chemsmart/pull/630/changes/057738a099c55757787508ed3be64f4426f9f868 so a direct call with constraints=None (or an empty list) is a no-op and no longer raises.

chemsmart/jobs/xtb/runner.py
Huiwen-Tan: yes, this mirrors the Gaussian/ORCA runner pattern, maybe can refactor this in the next PR.

xinglong-zhang (approved):
Things look great, thanks, @Huiwen-Tan!

PR #634, opened 2026-07-24
update RTD for GROUPER tool

ZihanL1:


xinglong-zhang:
updated in commit 2fcc6d9, as this one has fmt and linting errors.

PR #636, opened 2026-07-24
show grouper_RTD

ZihanL1:


ZihanL1:
fixed @xinglong-zhang

xinglong-zhang (approved):

xinglong-zhang:
Do you know what changed in ruff version 0.16.0 that is failing current tests? At some point, we will need to use newer versions of our dependencies.. @ZihanL1

xinglong-zhang:
According to GPT:

Ruff 0.16.0 enables 413 rules by default, compared with only 59 rules in 0.15.22. Your output contains newly default-enabled families such as UP, RUF, SIM, BLE, B, S, DTZ, TRY, PIE, PLW, and PERF. The code was already present; Ruff 0.15.22 simply did not check most of these rules. [Official Ruff 0.16 migration guide](https://astral.sh/blog/ruff-v0.16.0?utm_source=chatgpt.com)

The line:

Found 997 errors (458 fixed, 539 remaining).

also indicates that your lint command uses --fix. Ruff automatically fixed 458 findings but returned a nonzero exit code because 539 remained.

For an established project like CHEMSMART, I recommend preserving the previous lint policy first, then introducing additional rules gradually. Add this to `pyproject.toml:`

[tool.ruff.lint]
select = ["E4", "E7", "E9", "F"]

These are the pre-0.16 default rules. Retain any rules you were already selecting explicitly. For example, if you also require import sorting:

```
[tool.ruff.lint]
select = ["E4", "E7", "E9", "F", "I"]
```

This is better than simply pinning Ruff 0.15.22 because it makes your intended lint policy explicit and prevents future Ruff defaults from unexpectedly breaking CI.

I would not fix all 997 findings indiscriminately. Some are straightforward improvements, but others change design or behaviour:

UP031, SIM102, PIE810: mostly safe modernization.
RUF059: rename intentionally unused variables.
RUF012: usually annotate shared class constants with ClassVar; review whether mutation is intended.
BLE001, S110, S112: require deciding which exceptions should genuinely be suppressed.
TRY004: changes ValueError to TypeError, potentially affecting your public API and tests.
DTZ005/DTZ007: timezone changes may alter output semantics.
B019: may require redesigning caching rather than a mechanical edit.
N999: conflicts with your existing capitalized test filenames and can reasonably be excluded for tests.

A temporary alternative is pinning:

- ruff=0.15.22

**but explicitly setting lint.select is the more durable solution**. After that, you can add categories such as UP, B, SIM, and RUF individually and address them in manageable batches.

PR #640, opened 2026-07-24
update RTD for grouper(project setting,charge and multiplicity)

ZihanL1:


xinglong-zhang (changes requested):
no project setting is specified in bash commands - in these cases an error will be raised, as project setting is compulsory

ZihanL1:
added.

xinglong-zhang (approved):

PR #653, opened 2026-08-03
Add GROMACS smart presets

Ayson-HAN:
- Add high-level smart preset settings for GROMACS jobs
- Convert duration_ns to nsteps automatically
- Support protein, small-molecule, and membrane systems
- Use semiisotropic pressure coupling for membrane systems
- Preserve explicit user overrides and custom MDP files
- Add tests and an example project YAML


Ayson-HAN:
Closing this PR because the current branch does not yet contain the intended smart preset implementation. The feature will be revisited in a separate PR when it is ready.

PR #655, opened 2026-08-05
update RTD for configuration part

ZihanL1:
The original configuration commands are moved to 'getting started', so I rewrite the 'configuration-overview' and delete the 'configuration-test'


docs/source/configuration-test.rst
```diff
@@ -1,31 +0,0 @@
-#####################################
- Test Installation and Configuration
```
xinglong-zhang, line 2: so are these testing instructions somewhere in the RTD still?

docs/source/configuration-overview.rst
```diff
@@ -2,70 +2,15 @@
...
+modifying the **YAML** files in the ``~/.chemsmart/`` folder.
 
-****************************************
- For Linux, macOS, Ubuntu, and Clusters
-****************************************
+************
+ Next Steps
+************
 
-Run the configuration command:
```
xinglong-zhang, line 11: These are all deleted and I do not see those files in this PR, do they already exist:

-  :doc:`configuration-user-settings`
-  :doc:`configuration-server-settings`
-  :doc:`configuration-project-settings`

xinglong-zhang (changes requested):
@ZihanL1, this PR deletes information, but I cannot see these info added anywhere.

docs/source/configuration-test.rst
ZihanL1: image
This test command exists in 'getting started' part.

docs/source/configuration-overview.rst
ZihanL1: 1. 'make configure' is also finished in 'getting stated' part.
2. image
    They are hyperlinks directing users to the relevant documentation.

ZihanL1: Since the **Getting Started** guide already covers _make install, make configure, and installation testing_, I think this section should instead guide users through the additional configuration of the ~/.chemsmart/ directory.

PR #656, opened 2026-08-05
Fix: duplicate suffixes appended for ORCA jobs

liu-jing-yi:
In chemsmart/cli/orca/orca.py,when neither `-l` nor `-a` is given, ctx.invoked_subcommand (opt / sp / …) is appended twice:
```
    if label is None and append_label is None:
        label = os.path.splitext(os.path.basename(filename))[0]
        if filename:
            label = os.path.splitext(os.path.basename(filename))[0]
        else:
            label = "output"
        if ctx.invoked_subcommand:
            label = f"{label}_{ctx.invoked_subcommand}" # duplicate appending ctx.invoked_subcommand 
```

Also, Gaussian DB tags are wiped so default Gaussian DB jobs get bare `db_opt`, not `db_SID-…_opt` (ORCA/mol jobs keep the correct labels):

```
        if is_chemsmart_db:
            if structure_id is not None:
                label = f"{label}_SID-{structure_id}"
            elif record_id is not None:
                label = f"{label}_RID-{record_id}"
            elif record_index is not None:
                label = f"{label}_RI-{record_index}"
        if filename:
            label = os.path.splitext(os.path.basename(filename))[0]  # label is overwritten here.
        else:
            label = "output"

```
Both bugs were introduced accidentally by a merge commit e7773084, and is fixed in this PR. Thanks @Huiwen-Tan for spotting this!

### DIFF
--- a/chemsmart/cli/gaussian/gaussian.py
+++ b/chemsmart/cli/gaussian/gaussian.py
@@ -780,7 +780,10 @@ def gaussian(
                 label = f"{label}_RI-{record_index}"
         label = f"{label}_{append_label}"
     if label is None and append_label is None:
-        label = os.path.splitext(os.path.basename(filename))[0]
+        if filename:
+            label = os.path.splitext(os.path.basename(filename))[0]
+        else:
+            label = "output"
         if is_chemsmart_db:
             if structure_id is not None:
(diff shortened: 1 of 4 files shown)


xinglong-zhang (approved):
Thanks, @liu-jing-yi!

PR #676, opened 2026-08-21
fix: CI error triggered by pymol import

liu-jing-yi:
## **Summary**
Stop CHEMSMART from requiring PyMOL just to import. `zhang_group_scientific_styles.py` used to import pymol at module load, so any import of CHEMSMART failed in environments (e.g., the CI workflow environment) without PyMOL, even when visualization was unused.

PyMOL is now an optional import, so packages like Conflow can import CHEMSMART without it.

## **Changes**
Wrap `from pymol import cmd` in a `try/except ImportError` block and `set cmd = None` when PyMOL is missing.


xinglong-zhang (approved):
Thanks, @liu-jing-yi

PR #686, opened 2026-09-10
Add native ORCA MECP calculations

TIntt0:
Add MECP job settings and CLI support for two-state multiplicities, broken symmetry, solvent.
Validate multiplicities and broken-symmetry compatibility before execution.
Parse crossing-point geometry, state energies, energy-gap history, and both PES frequency tables.
Generate quality reports covering normal termination, optimization convergence, energy-gap tolerance, and imaginary frequencies.
Require complete, consecutively indexed 3N frequency tables before marking NumFreq complete; incomplete results remain unclassified.
Add regression coverage using saved FeO⁺, CH₃O/CH₂OH, twisted-ethylene, CO-pathway, and OOH-pathway outputs, including warning and passing cases.


chemsmart/jobs/orca/mecp.py
```diff
@@ -0,0 +1,221 @@
...
+                f"{broken_sym_multiplicity}, which is incompatible with "
+                f"the molecule's {electron_count} electrons."
+            )
+
+    @property
+    def report_file(self):
+        """Path to the concise, user-facing MECP quality report."""
+        return os.path.join(self.folder, f"{self.label}_report.log")
+
+    def write_report(self, energy_gap_tolerance=None):
```
xinglong-zhang, line 114: This was called log_result in Gaussian but write_report in ORCA? We may need to standardise.

chemsmart/jobs/orca/settings.py
```diff
@@ -2132,3 +2132,150 @@ def _get_neb_route_string(self):
...
+        maxiter (int | None): Maximum number of SurfCrossOpt iterations
+            (written as ``MaxIter`` in ``%geom``; the ORCA ``%mecp`` block
+            has no ``MaxIter`` keyword).
+    """
+
+    MODES = frozenset({"opt", "numfreq"})
+
+    def __init__(
+        self,
+        multiplicity_a=None,
```
xinglong-zhang, line 2162: use multiplicity1 and multiplicity2 for consistency

xinglong-zhang (changes requested):
Similar fixes required for Gaussian MECP. @TIntt0

chemsmart/jobs/orca/mecp.py
```diff
@@ -0,0 +1,221 @@
...
+                f"{broken_sym_multiplicity}, which is incompatible with "
+                f"the molecule's {electron_count} electrons."
+            )
+
+    @property
+    def report_file(self):
+        """Path to the concise, user-facing MECP quality report."""
+        return os.path.join(self.folder, f"{self.label}_report.log")
+
+    def log_result(self, energy_gap_tolerance=None):
```
xinglong-zhang, line 114: In MECP for Gaussian, you used `log_step()`, here you used `log_result()`, are they meant to be different? If not, it's better to use one standardised one, since then it's easier to remember and use elsewhere.

`MECPSoftwareJob.log_results()` which will be the same for any software implementing MECP job.

xinglong-zhang (changes requested):
Looking great, just one standardisation to consider, @TIntt0

PR #687, opened 2026-09-10
Add Gaussian-MECP optimization with Harvey BFGS

TIntt0:
This PR improves the Gaussian MECP workflow with a more robust optimization implementation, restart support, safer checkpoint handling, clearer CLI options, and expanded tests and documentation.

The Harvey inverse-BFGS optimizer is now the default MECP optimization method. The existing Barzilai–Borwein and grow/shrink methods remain available for comparison and fallback use.

### DIFF
--- a/chemsmart/cli/chemsmart
+++ b/chemsmart/cli/chemsmart
@@ -1,0 +1,9 @@
+#!/usr/bin/env python
+import re
+import sys
+
+from chemsmart.cli.main import main
+
+if __name__ == '__main__':
+    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
+    sys.exit(main())
--- a/chemsmart/cli/chemsmart
+++ b/chemsmart/cli/chemsmart
@@ -1,9 +1,0 @@
-#!/usr/bin/env python
-import re
-import sys
-
-from chemsmart.cli.main import main
-
-if __name__ == '__main__':
-    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
-    sys.exit(main())
--- a/chemsmart/cli/gaussian/link.py
+++ b/chemsmart/cli/gaussian/link.py
@@ -8,11 +8,15 @@ from chemsmart.cli.gaussian.gaussian import (
     click_gaussian_solvent_options,
     gaussian,
 )
+<<<<<<< Updated upstream
 from chemsmart.cli.gaussian.mecp_options import (
     add_mecp_method_suffix,
     click_mecp_restart_option,
     click_mecp_step_size_method_option,
 )
+=======
+from chemsmart.cli.gaussian.mecp_options import click_gaussian_mecp_state_options
+>>>>>>> Stashed changes
 from chemsmart.cli.job import click_job_options
 from chemsmart.utils.cli import (
     MyCommand,
@@ -49,6 +53,7 @@ logger = logging.getLogger(__name__)
     "--route", type=str, default=None, help="Route for the link section."
 )
 # MECP-specific options (used when --jobtype mecp)
+<<<<<<< Updated upstream
 @click.option(
     "--multiplicity-1",
     "-m1",
@@ -81,6 +86,9 @@ logger = logging.getLogger(__name__)
     default=None,
     help="[MECP] Charge for state 2. Defaults to charge-1.",
 )
+=======
+@click_gaussian_mecp_state_options
+>>>>>>> Stashed changes
 @click.option(
     "--max-steps",
     type=int,
@@ -400,7 +408,11 @@ def _link_mecp(
     if mecp_settings.multiplicity_a is None:
         raise ValueError(
             "State A multiplicity is not set. "
+<<<<<<< Updated upstream
             "Use gaussian -m/--multiplicity or link -j mecp --multiplicity-1/-m1."
+=======
+            "Use gaussian -m/--multiplicity or link -j mecp --multiplicity1."
+>>>>>>> Stashed changes
         )
 
     if multiplicity2 is None:
@@ -415,7 +427,11 @@ def _link_mecp(
     if mecp_settings.charge_a is None:
         raise ValueError(
             "State A charge is not set. "
+<<<<<<< Updated upstream
             "Use gaussian -c/--charge or link -j mecp --charge-1/-c1."
+=======
+            "Use gaussian -c/--charge or link -j mecp --charge1."
+>>>>>>> Stashed changes
         )
 
     if charge2 is None:
--- a/chemsmart/cli/gaussian/link.py
+++ b/chemsmart/cli/gaussian/link.py
@@ -124,16 +124,15 @@ logger = logging.getLogger(__name__)
 @click.option(
     "--step-size-method",
     type=click.Choice(
-        ["bb", "grow_shrink", "harvey", "harvey_bfgs"],
+        ["bb", "grow_shrink", "harvey"],
         case_sensitive=False,
     ),
     default=None,
     help=(
         "[MECP] Step size adaptation algorithm when adaptive is enabled. "
-        "'bb' uses the Barzilai-Borwein secant rule (default). "
-        "'grow_shrink' uses a merit-based grow/shrink rule; 'harvey' uses "
-        "Harvey's energy-based heuristic; 'harvey_bfgs' uses the full "
-        "Harvey quasi-Newton optimizer."
+        "'harvey' uses the Harvey inverse-BFGS optimizer (default). "
+        "'bb' uses the Barzilai-Borwein secant rule; 'grow_shrink' uses "
+        "a merit-based grow/shrink rule."
     ),
 )
 @click.option(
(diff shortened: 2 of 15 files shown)


chemsmart/cli/gaussian/link.py
```diff
@@ -45,28 +50,36 @@
 )
 # MECP-specific options (used when --jobtype mecp)
 @click.option(
-    "--multiplicity-a",
+    "--multiplicity-1",
```
xinglong-zhang, line 53: use `--multiplicity1` instead of `--multiplicity-1`; the former will make help messages multiplicity1 + 2 easier to understand.

chemsmart/cli/gaussian/link.py
```diff
@@ -45,28 +50,36 @@
...
+    "-m1",
+    "multiplicity1",
     type=int,
     default=None,
-    help="[MECP] Spin multiplicity for state A.",
+    help="[MECP] Spin multiplicity for state 1.",
 )
 @click.option(
-    "--multiplicity-b",
+    "--multiplicity-2",
```
xinglong-zhang, line 61: similarly, use `--multiplicity2` instead.

chemsmart/cli/gaussian/link.py
```diff
@@ -45,28 +50,36 @@
...
+    "-m2",
+    "multiplicity2",
     type=int,
     default=None,
-    help="[MECP] Spin multiplicity for state B. Defaults to multiplicity-a + 2.",
+    help="[MECP] Spin multiplicity for state 2. Defaults to multiplicity-1 + 2.",
 )
 @click.option(
-    "--charge-a",
+    "--charge-1",
```
xinglong-zhang, line 69: use `--charge1`

chemsmart/cli/gaussian/link.py
```diff
@@ -45,28 +50,36 @@
...
+    "-c1",
+    "charge1",
     type=int,
     default=None,
-    help="[MECP] Charge for state A.",
+    help="[MECP] Charge for state 1.",
 )
 @click.option(
-    "--charge-b",
+    "--charge-2",
```
xinglong-zhang, line 77: use `--charge2`

chemsmart/cli/gaussian/link.py
```diff
@@ -45,28 +50,36 @@
 )
 # MECP-specific options (used when --jobtype mecp)
 @click.option(
```
xinglong-zhang, line 52: Seems that these click options can be wrapped as a decorator that can be reused.

chemsmart/cli/gaussian/mecp.py
```diff
@@ -14,28 +19,36 @@
 @gaussian.command("mecp", cls=MyCommand)
 @click_job_options
 @click.option(
```
xinglong-zhang, line 21: Wrap as a reusable decorator.

chemsmart/cli/chemsmart
xinglong-zhang, line 1: Why is there a file mode change here? Changed from 755 to 644?

chemsmart/io/gaussian/output.py
```diff
@@ -173,6 +173,20 @@ def spin(self):
                 return spin
         return None
 
+    @property
+    def spin_squared(self):
+        """Return the final post-annihilation ``<S^2>`` value, if printed."""
+        pattern = re.compile(
```
xinglong-zhang, line 179: may need to separate into two properties (before and after)

chemsmart/io/gaussian/output.py
```diff
@@ -173,6 +173,20 @@ def spin(self):
...
+            r"S\*\*2 before annihilation\s+[-+0-9.DEde]+,\s+after\s+"
```
xinglong-zhang, line 180: regex pattern may be placed in utils.repatterns.py

chemsmart/jobs/gaussian/job.py
```diff
@@ -138,7 +138,7 @@ def chkfile(self):
         Returns:
             str: Full path to the Gaussian checkpoint file.
         """
-        chkfile = self.label + ".chk"
+        chkfile = getattr(self, "checkpoint_filename", self.label + ".chk")
```
xinglong-zhang, line 141: Why use getattr()? There may not be an attribute named "checkpoint_filename".

xinglong-zhang: I guess you probably need 
```
if self.label is None:
    <...>
else:
    <...>
```

chemsmart/jobs/gaussian/mecp.py
```diff
@@ -44,6 +44,40 @@ class GaussianMECPJob(GaussianJob):
     TYPE = "g16mecp"
     MIN_DIFF_GRAD_NORM_SQ = 1.0e-20
 
+    @staticmethod
+    def _without_unavailable_guess_read(route):
```
xinglong-zhang, line 48: Is there special meaning for this method name? why not use `_with_available_guess_read`?

chemsmart/jobs/gaussian/mecp.py
```diff
@@ -44,6 +44,40 @@ class GaussianMECPJob(GaussianJob):
...
+        if re.search(r"\bsymm(?:etry)?\b", route, flags=re.IGNORECASE):
+            raise ValueError(
+                "MECP requires nosymm so Cartesian forces remain aligned with "
+                "the optimization coordinates; remove the explicit symmetry option."
+            )
+        return f"{route} nosymm".strip()
+
     # MECP-specific attribute names that must be stripped when building
     # a GaussianLinkJobSettings for each sub-job (broken-symmetry mode).
     _MECP_ONLY_KEYS = frozenset(
```
xinglong-zhang, line 83: probably best to be placed at the beginning of the file

chemsmart/jobs/gaussian/mecp.py
```diff
@@ -275,19 +425,206 @@ def _mecp_displacement(self, energy_diff, grad_a, grad_b, step_size):
...
+            par_grad,
+            seam_correction,
+            inv_hess,
+            update_status,
+            fac,
+            fae,
+            eff_grad,
+        )
+
+    _GROW_SHRINK_IMPROVEMENT_THRESHOLD = 0.10
```
xinglong-zhang, line 605: can these be collected as a dict for grow_shrink and another dict for BB?

xinglong-zhang: and place as global keys similar to _MECP_ONLY_KEYS?

chemsmart/jobs/gaussian/mecp.py
```diff
@@ -365,7 +724,15 @@ def _is_converged(self, energy_diff, eff_grad, displacement):
         )
 
     def _log_step(
```
xinglong-zhang, line 726: maybe we should change this into a public method, if an MECP job object can be passed to other class, then mecp_job.log_step() might be useful.

chemsmart/jobs/gaussian/runner.py
```diff
@@ -98,6 +98,17 @@ class GaussianJobRunner(JobRunner):
     # of instance attribute so it needs not be set at
     # instance level - set during initialization (__init__).
 
+    def _scratch_job_directory(self, job):
+        """Return the scratch directory, optionally grouped by a job folder."""
+        scratch_parent = getattr(job, "scratch_parent_folder", None)
```
xinglong-zhang, line 103: scratch_parent_folder attribute is not created for the job? I prefer that usable attributes be created in the class, rather than at runtime (due to the fact that if the latter, one may not be able to access that attribute as you don't know if the function creating the attribute is run or not), thus I will usually avoid using getattr() but instead have the attribute set up in the class constructor.

chemsmart/jobs/gaussian/runner.py
```diff
@@ -98,6 +98,17 @@ class GaussianJobRunner(JobRunner):
...
+        if scratch_parent:
+            return os.path.join(
+                self.scratch_dir,
+                os.path.basename(scratch_parent),
+                job.label,
+            )
+        return os.path.join(self.scratch_dir, job.label)
```
xinglong-zhang, line 110: If self.scratch_dir is None, wouldn't this cause error?

docs/source/gaussian-other-jobs.rst
```diff
@@ -66,24 +66,24 @@ MECP Options
       -  Default
       -  Description
 
-   -  -  ``--multiplicity-a``
+   -  -  ``--multiplicity-1`` / ``-m1``
```
xinglong-zhang, line 69: to be updated accordingly

xinglong-zhang (changes requested):
Great work and you have achieved so much in a month+!! Some comments for consideration, @TIntt0

chemsmart/jobs/gaussian/mecp.py
```diff
@@ -234,6 +386,11 @@ def _run_state(self, positions_bohr, step_idx, state):
                 f"coordinate shape {np.array(mol.positions).shape}."
             )
         gradient = -forces
+        if not hasattr(self, "_last_spin_squared"):
+            self._last_spin_squared = {"A": None, "B": None}
```
xinglong-zhang, line 390: Can this be initialised in the object constructor, rather than has it checked and set at run time?

chemsmart/cli/chemsmart
xinglong-zhang: Is this file permission changed?

xinglong-zhang (changes requested):
Just two comments and then we can merge, @TIntt0

PR #689, opened 2026-09-11
Fix custom solv problem for gaussian jobs

ZihanL1:
When a Gaussian optimization uses custom SMD solvent parameters, using its **.log file** as input for another custom-SMD single-point calculation, eg.

`chemsmart sub -s cu_new gaussian -p outer_solv -f xxx_opt.log(use custom SMD solvent parameters in optimization) sp`

it causes an error:

`ValueError: Custom solvent parameters must be a string!`

CHEMSMART should read the optimized geometry and override the old solvent settings with those from **outer_solv**

This PR should have fixed this problem (already tested on cluster)


xinglong-zhang (approved):
Thanks, @ZihanL1!

PR #695, opened 2026-09-16
feat(orca): add TDDFT jobs with SOC and NTO support

Lewen-WangCC:
Add CLI options, settings, and input blocks for vertical TDDFT/TDA and opt-in excited-state optimization and frequencies.

Preserve additional route parameters, document TD workflows, and add regression coverage. Stop scratch cleanup when output recovery fails.


chemsmart/cli/orca/td.py
```diff
@@ -0,0 +1,446 @@
...
+                param_hint="'--root'",
+            )
+        if nroots is not None and effective_root > nroots:
+            raise click.UsageError(
+                f"--root ({effective_root}) exceeds --nroots ({nroots}); "
+                "increase --nroots or lower --root."
+            )
+        if not excited_task and root is None:
+            # Keep IRoot off for pure vertical TD unless the user opted in.
+            effective_root = None
```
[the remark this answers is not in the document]
Lewen-WangCC: deleted

xinglong-zhang:
> ### 🔵 Needs a closer look
> It is a large new feature that also changes ORCA scratch-recovery behavior for all job types (`_postrun` now re-raises), which has broad, untested operational impact warranting human review.
> 
> Review details

@Lewen-WangCC, thanks for introducing this feature, please see if this comment needs to be addressed.

Lewen-WangCC:
The ”raise“ added in chemsmart/jobs/orca/runner.py, line 460 ‎prevents scratch cleanup when an output file cannot be copied back, preserving unrecovered results for manual recovery. I have encountered a similar situation where the job disappeared, the scratch files were deleted, and the calculation files were never copied back to the working directory. This is difficult to reproduce and test in practice, so for now, I can only verify that the code logic is sound.

The "raise" at the end of _post_run() will only prevents the subsequent _postrun_cleanup() call.

As for the help info problem, solved. @xinglong-zhang

Lewen-WangCC:
This branch has been tested on the hpc, using command "chemsmart sub --delete-scratch -s cu orca -p ljb -f s1_cf3_me_ts2_c0m1ts_ircf_opt.log -c 0 -m 1 -d DEFGRID3 td --nroots 10 --triplets --no-tda --dosoc --printlevel 3 --no-cpcmeq --soc-type 3", inp written as expected 
image
This job terminated normally.

chemsmart/cli/orca/tddft.py
```diff
@@ -0,0 +1,444 @@
+"""ORCA TDDFT / TDA CLI command."""
```
xinglong-zhang, line 1: cli is td, but here is tddft, we may need to standardize. Gaussian has cli.gaussian.tddft.py and job.gaussian.tddft.py while the command is td (for simplicity at job run).

Perhaps here it can be cli.orca.tddft.py too for file name

tests/test_ORCATDDFT.py
```diff
@@ -0,0 +1,1368 @@
...
+
+from chemsmart.cli.orca.orca import orca as orca_cli
+from chemsmart.jobs.orca import ORCATDDFTJob
+from chemsmart.jobs.orca.settings import ORCATDDFTJobSettings
+from chemsmart.jobs.orca.tddft import ORCATDDFTJob as _ORCATDDFTJob_alias
+from chemsmart.jobs.orca.writer import ORCAInputWriter
+from chemsmart.settings.orca import ORCAProjectSettings
+
+
+@pytest.fixture
```
xinglong-zhang: Please move all fixtures to conftest.py.

xinglong-zhang (changes requested):
Thanks, @Lewen-WangCC, some comments for your consideration.

tests/test_ORCATDDFT.py
Lewen-WangCC: solved

chemsmart/cli/orca/tddft.py
Lewen-WangCC: solved

xinglong-zhang (approved):

PR #697, opened 2026-09-18
Distance as starting point for collective variables

laurag172213-bit:
1. Add Gaussian recorded-IRC geometry extraction and distance-analysis functions. 
2. Add single-point analysis, comparison between two selected points, and full-IRC summary matrices containing absolute start-to-end changes and maximum–minimum ranges.
3. Includes optional connectivity-based pair selection using ChemSmart’s existing method, while retaining all-pair reference results.


xinglong-zhang (changes requested):
I think in general this is not quite right, each Molecule object should have the distances, angles and dihedrals matrix. The IRC trajectory/output file can already extract the Molecules object, so it is needed that each Molecule object have these matrices. As discussed, there should be Molecule.distances_matrix, Molecule.angles_matrix and Molecule.dihedrals_matrix properties. @laurag172213-bit

laurag172213-bit:
I updated the PR to add distance, angle and dihedral matrix properties to Molecule, together with trajectory summaries using the existing reader and I tried to test it with six previous irc.logs. I have removed last week's IRC-specific code, but I would like to clarify why I originally added a separate extraction function: Gaussian IRC output can include correction-attempt geometries that should not be treated as accepted IRC points. Please see the attached document for a detailed explanation and example. If this needs to be addressed for our analysis, I could try adding a Gaussian-specific filter while keeping the existing reader unchanged.
[Why use IRC-specific extraction for CV analysis.docx](https://github.com/user-attachments/files/32463780/Why.use.IRC-specific.extraction.for.CV.analysis.docx)

